PR #1, opened 2015-11-10
Proof of Concept

edwardloveall:
A redirection has:

* A slug (to find with)
* A url (to redirect to)
* A next (reference to another redirection)
* A previous (virtual reference to another redirection)

Using this, we can do `redirection.next.url` and `redirection.previous.url`

This is, as the title suggests a proof of concept. There's a bit of cleaning up (using the slug as a default find method, law of demeter in controller) to do. Lemme know what cha think.

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -1,0 +1,13 @@
+class RedirectionsController < ApplicationController
+  def next
+    redirection = Redirection.find_by(slug: params[:slug])
+
+    redirect_to redirection.next_url
+  end
+
+  def previous
+    redirection = Redirection.find_by(slug: params[:slug])
+
+    redirect_to redirection.previous_url
+  end
+end
--- a/app/models/redirection.rb
+++ b/app/models/redirection.rb
@@ -1,0 +1,16 @@
+class Redirection < ActiveRecord::Base
+  belongs_to :next, class_name: "Redirection"
+  has_one :previous, class_name: "Redirection", foreign_key: "next_id"
+
+  validates :next_id, uniqueness: true
+  validates :slug, presence: true, uniqueness: true
+  validates :url, presence: true
+
+  def next_url
+    self.next.url
+  end
+
+  def previous_url
+    previous.url
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,2 +1,4 @@
 Rails.application.routes.draw do
+  get ":slug/next", to: "redirections#next"
+  get ":slug/previous", to: "redirections#previous"
 end
--- a/db/migrate/20151110191852_create_redirections.rb
+++ b/db/migrate/20151110191852_create_redirections.rb
@@ -1,0 +1,10 @@
+class CreateRedirections < ActiveRecord::Migration
+  def change
+    create_table :redirections do |t|
+      t.timestamps null: false
+      t.references :next
+      t.string :slug, index: true
+      t.text :url
+    end
+  end
+end
--- a/db/migrate/20151110191852_create_redirections.rb
+++ b/db/migrate/20151110191852_create_redirections.rb
@@ -2,9 +2,13 @@ class CreateRedirections < ActiveRecord::Migration
   def change
     create_table :redirections do |t|
       t.timestamps null: false
-      t.references :next
-      t.string :slug, index: true
-      t.text :url
+      t.integer :next_id, null: false
+      t.string :slug, null: false
+      t.text :url, null: false
     end
+    add_foreign_key :redirections, :redirections, column: :next_id
+    add_index :redirections, :next_id, unique: true
+    add_index :redirections, :slug, unique: true
+    add_index :redirections, :url, unique: true
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -35,11 +35,14 @@ ActiveRecord::Schema.define(version: 20151110191852) do
   create_table "redirections", force: :cascade do |t|
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.integer  "next_id"
-    t.string   "slug"
-    t.text     "url"
+    t.integer  "next_id",    null: false
+    t.string   "slug",       null: false
+    t.text     "url",        null: false
   end
 
-  add_index "redirections", ["slug"], name: "index_redirections_on_slug", using: :btree
+  add_index "redirections", ["next_id"], name: "index_redirections_on_next_id", unique: true, using: :btree
+  add_index "redirections", ["slug"], name: "index_redirections_on_slug", unique: true, using: :btree
+  add_index "redirections", ["url"], name: "index_redirections_on_url", unique: true, using: :btree
 
+  add_foreign_key "redirections", "redirections", column: "next_id"
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151103053545) do
+ActiveRecord::Schema.define(version: 20151110191852) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -31,5 +31,15 @@ ActiveRecord::Schema.define(version: 20151103053545) do
   end
 
   add_index "delayed_jobs", ["priority", "run_at"], name: "delayed_jobs_priority", using: :btree
+
+  create_table "redirections", force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.integer  "next_id"
+    t.string   "slug"
+    t.text     "url"
+  end
+
+  add_index "redirections", ["slug"], name: "index_redirections_on_slug", using: :btree
 
 end
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe RedirectionsController do
+RSpec.describe RedirectionsController do
   describe "GET :next" do
     it "redirects to the next site" do
       gabe = create(:redirection, :gabe)
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe RedirectionsController do
       edward = create(:redirection, :edward, next: gabe)
       gabe.update(next: edward)
 
-      get :next, slug: "gabe"
+      get :next, slug: gabe.slug
 
       expect(response).to redirect_to(edward.url)
     end
@@ -19,7 +19,7 @@ RSpec.describe RedirectionsController do
       edward = create(:redirection, :edward, next: gabe)
       gabe.update(next: edward)
 
-      get :previous, slug: "edward"
+      get :previous, slug: edward.slug
 
       expect(response).to redirect_to(gabe.url)
     end
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+describe RedirectionsController do
+  describe "GET :next" do
+    it "redirects to the next site" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      get :next, slug: "gabe"
+
+      expect(response).to redirect_to(edward.url)
+    end
+  end
+
+  describe "GET :previous" do
+    it "redirects to the previous site" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      get :previous, slug: "edward"
+
+      expect(response).to redirect_to(gabe.url)
+    end
+  end
+end
--- a/spec/models/redirection_spec.rb
+++ b/spec/models/redirection_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe Redirection do
+RSpec.describe Redirection do
   describe "associations" do
     it { should belong_to(:next).
                 class_name("Redirection") }
--- a/spec/models/redirection_spec.rb
+++ b/spec/models/redirection_spec.rb
@@ -1,0 +1,38 @@
+require "rails_helper"
+
+describe Redirection do
+  describe "associations" do
+    it { should belong_to(:next).
+                class_name("Redirection") }
+    it { should have_one(:previous).
+                class_name("Redirection").
+                with_foreign_key("next_id") }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of :slug }
+    it { should validate_presence_of :url }
+    it { should validate_uniqueness_of :next_id }
+    it { should validate_uniqueness_of :slug }
+  end
+
+  describe "#next_url" do
+    it "returns the url of the next referenced redirection" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      expect(gabe.next_url).to eq(edward.url)
+    end
+  end
+
+  describe "#previous_url" do
+    it "returns the url of the previous referenced redirection" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      expect(gabe.previous_url).to eq(edward.url)
+    end
+  end
+end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -18,6 +18,7 @@ RSpec.configure do |config|
 
   config.example_status_persistence_file_path = "tmp/rspec_examples.txt"
   config.order = :random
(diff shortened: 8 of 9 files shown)


db/migrate/20151110191852_create_redirections.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRedirections < ActiveRecord::Migration
+  def change
+    create_table :redirections do |t|
+      t.timestamps null: false
+      t.references :next
+      t.string :slug, index: true
+      t.string :url, length: 255
```
gabebw: Why have a length limit?

spec/controllers/redirections_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+describe RedirectionsController do
+  describe "GET :next" do
+    it "redirects to the next site" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      get :next, slug: "gabe"
+
+      expect(response).to redirect_to("http://edwardloveall.com")
```
gabebw: Maybe "redirect_to edward.url"?

config/routes.rb
```diff
@@ -1,2 +1,4 @@
 Rails.application.routes.draw do
+  get ":slug/next", to: "redirections#next"
```
gabebw: What are the advantages to having a custom slug versus using the referrer and `/next`?

app/controllers/redirections_controller.rb
```diff
@@ -0,0 +1,13 @@
+class RedirectionsController < ApplicationController
+  def next
+    redirection = Redirection.find_by(slug: params[:slug])
```
gabebw: If a slug doesn't exist, this won't create it -- is that intentional? I was thinking something like this would be cool:

* Anyone adds a `/next` link
* When someone clicks that link we create a record
* No registration required

db/migrate/20151110191852_create_redirections.rb
edwardloveall: Maybe i'm mis-remembering, but I thought the default length of a string was 128 char? Can't find it now, but if so, this will make sure it's longer than that because we can has some whack-ass long urls.

gabebw: > we can has some whack-ass long urls.

What if we bypass that entirely and use a `text` instead of a `string`? Then we could put up to 2GB of text in that column.

gabebw: Oops, 1GB: http://tinyrobots.thoughtbot.com/post/6293521440/joe-lays-down-the-law

config/routes.rb
edwardloveall: It decouples the site that it came from the site that it's going to, which could also be a security disadvantage, i.e. anyone could use someones slug and start the webring. Although I dunno what the advantage of that would be.

Also, this allows people to have multiple subdomains that hook into the webring, i.e. `example.com` and `blog.example.com` could both use the `example` slug, instead of having multiple sites listed for each redirection or multiple sites from the same domain in the ring.

Also, we don't have to parse URLs either, even though that seems pretty solid, there could be weird edge cases.

Seems more flexible, if not as slick with the same link for everyone.

spec/models/redirection_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+                class_name("Redirection") }
+    it { should have_one(:previous).
+                class_name("Redirection").
+                with_foreign_key("next_id") }
+  end
+
+  describe "validations" do
+    it { should validate_presence_of :slug }
+    it { should validate_presence_of :url }
+    it { should validate_uniqueness_of(:next_id) }
```
gabebw: Why parentheses here but not in the others?

db/migrate/20151110191852_create_redirections.rb
edwardloveall: Thought about that too but it seemed excessive. Honestly though, this is a webring so...

spec/controllers/redirections_controller_spec.rb
edwardloveall: Yeah this is more brittle but everytime I write expectations with variables they end up too clever.

config/routes.rb
gabebw: > Also, this allows people to have multiple subdomains that hook into the webring, i.e. example.com and blog.example.com could both use the example slug, instead of having multiple sites listed for each redirection or multiple sites from the same domain in the ring.

Duh, I forgot about that. Cool.

app/controllers/redirections_controller.rb
edwardloveall: Hmmm interesting, I had the thought that we would be registering them and setting the order. This is a whole new twist...

spec/models/redirection_spec.rb
edwardloveall: Oh whoops, I as doing `allow_blank` but it's [broken](https://github.com/thoughtbot/shoulda-matchers/issues/776)

app/controllers/redirections_controller.rb
edwardloveall: So is this what you're picturing:

Current webring:

```
gabebw.com         added: 10.31.2015 ⤓
alexfox.me         added: 11.02.2015 ⤴
adarsh.io          added: 11.03.2015 ⤴
```

(`⤴` represents next link direction, `⤓` to the bottom)

If I link to the webring, it would look for my slug or referring URL (however we architect it), and create it if it didn't exist. Then it would tack me on to the end, pointing toward the last known site in the ring. So everything would then look like:

```
gabebw.com         added: 10.31.2015 ⤓
alexfox.me         added: 11.02.2015 ⤴
adarsh.io          added: 11.03.2015 ⤴
edwardloveall.com  added: 11.07.2015 ⤴
```

Is that right?

spec/controllers/redirections_controller_spec.rb
adarsh: +1 for the variable in the expectation, the 'less brittle' part is significant.

app/controllers/redirections_controller.rb
gabebw: Yup, that's what I'm thinking. If we need to we can always deregister or blacklist or whitelist people.

adarsh:
Solid commit messaging. It's shenanigans like that which keep side projects alive.

app/controllers/redirections_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    redirect_to redirection.next.url
+  end
+
+  def previous
```
gabebw: I don't love not using RESTful actions, but hey: Proof Of Concept! Proof Of Concept! WOOOOO

db/migrate/20151110191852_create_redirections.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRedirections < ActiveRecord::Migration
+  def change
+    create_table :redirections do |t|
+      t.timestamps null: false
+      t.references :next
```
gabebw: `null: false` for everything?

app/controllers/redirections_controller.rb
adarsh: Or 'brownlist'.

db/migrate/20151110191852_create_redirections.rb
gabebw: Actually I'd do this for this line:

```ruby
t.references :next, index: true, foreign_key: true, null: false
```

That way we get an index AND a foreign key constraint.

gabebw: (Why an index? Because it references another table, and an index will speed up Rails' querying.)

spec/controllers/redirections_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  describe "GET :previous" do
+    it "redirects to the previous site" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      get :previous, slug: "edward"
+
+      expect(response).to redirect_to("http://gabebw.com")
```
gabebw: `gabe.url`

app/controllers/redirections_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    redirection = Redirection.find_by(slug: params[:slug])
+
+    redirect_to redirection.previous.url
```
adarsh: What about creating a method on `Redirection` to avoid this Demeter violation?

```ruby
# redirection.rb
# ...
  def previous_url
    previous.url
  end
```

db/schema.rb
```diff
@@ -32,4 +32,14 @@
 
   add_index "delayed_jobs", ["priority", "run_at"], name: "delayed_jobs_priority", using: :btree
 
+  create_table "redirections", force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.integer  "next_id"
```
adarsh: Index foreign keys?

app/controllers/redirections_controller.rb
edwardloveall: Yeah, I just don't know what it would be though. A NextsController#show action and a PreviousesController#show ?

Maybe a `NavigationsController#show` with `params[:direction]`?

We're not really dealing with traditional crud resources. All for changing it up though if you have a good naming convention in mind.

gabebw: > We're not really dealing with traditional crud resources.

We're dealing with throwback REST resources: REST only dates from 2000, while webrings date from like 1995.

This is fine.

spec/controllers/redirections_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+describe RedirectionsController do
```
adarsh: A nit, but the preferred new syntax is `Rspec.describe BlahClassName do` ([link to docs](https://relishapp.com/rspec/rspec-core/v/3-0/docs/configuration/global-namespace-dsl)).

db/migrate/20151110191852_create_redirections.rb
edwardloveall: I'm not a big fan of database validations. They explode pretty hard. I know there are differing schools of thought, but I find AR validations much easier to deal with and just as powerful for anything we'd likely need here. The timestamps one was just in there.

spec/controllers/redirections_controller_spec.rb
gabebw: EXTRA nitpick it is `RSpec.describe` and yeah let's do it, let's go Full RSpec

app/controllers/redirections_controller.rb
edwardloveall: Yup. Totally planned.

spec/controllers/redirections_controller_spec.rb
edwardloveall: eeeeinteresting. I did not know this was a thing now. I certainly don't like money patching. Perhaps I'll give this a shot on this project.

db/migrate/20151110191852_create_redirections.rb
edwardloveall: Does it matter that it doesn't reference another table but another record in this table? Probably not, I just have a very rudimentary understanding of indexes

gabebw: Yeah that's cool. Oh hmm, actually -- should we have uniqueness validations on url, and next so that two redirections don't have the same URL and don't point at the same place?

spec/controllers/redirections_controller_spec.rb
adarsh: I love MONEY PATCHING.

[image: Inline image 1]



-- 
313.454.1515 (mobile)

db/migrate/20151110191852_create_redirections.rb
gabebw: I like DB validations for uniqueness because they actually provide a service: AR uniqueness validations can have race conditions.

gabebw: > Does it matter that it doesn't reference another table but another record in this table? Probably not, I just have a very rudimentary understanding of indexes

I think an index can't _hurt_

adarsh: Also you can sneak around AR validations in the console. The DB is like the
Honey Badger in it's level of regard for others.



-- 
313.454.1515 (mobile)

spec/controllers/redirections_controller_spec.rb
gabebw: "Inline image 1" is straight up _tantalizing_

adarsh: Wow, thanks `GitHubEmailParser`.

Let's try again:

I love MONEY PATCHING.

money patching

app/controllers/redirections_controller.rb
edwardloveall: Gonna work on this in another branch

db/migrate/20151110191852_create_redirections.rb
edwardloveall: Okay. DB validations it is

app/controllers/redirections_controller.rb
gabebw: :+1:

db/migrate/20151110191852_create_redirections.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :redirections do |t|
+      t.timestamps null: false
+      t.integer :next_id, null: false
+      t.string :slug, null: false
+      t.text :url, null: false
+    end
+    add_foreign_key :redirections, :redirections, column: :next_id
+    add_index :redirections, :next_id, unique: true
+    add_index :redirections, :slug, unique: true
+    add_index :redirections, :url, unique: true
```
edwardloveall: I think this is the right idea, yea?

gabebw: Yup!

spec/spec_helper.rb
```diff
@@ -18,6 +18,7 @@
 
   config.example_status_persistence_file_path = "tmp/rspec_examples.txt"
   config.order = :random
+  config.expose_dsl_globally = false
```
gabebw: Nice.

gabebw:
This looks great. :+1:

edwardloveall:
Cool. Gonna squash down some of these commits and then :ship:

app/models/redirection.rb
```diff
@@ -0,0 +1,16 @@
+class Redirection < ActiveRecord::Base
+  belongs_to :next, class_name: "Redirection"
+  has_one :previous, class_name: "Redirection", foreign_key: "next_id"
+
+  validates :next_id, uniqueness: true, allow_blank: true
+  validates :slug, presence: true, uniqueness: true
+  validates :url, presence: true
+
+  def next_url
+    self.next.url
```
gabebw: Is the `self` here and in `previous_url` necessary?

app/models/redirection.rb
```diff
@@ -0,0 +1,16 @@
+class Redirection < ActiveRecord::Base
+  belongs_to :next, class_name: "Redirection"
+  has_one :previous, class_name: "Redirection", foreign_key: "next_id"
+
+  validates :next_id, uniqueness: true, allow_blank: true
```
gabebw: Hmm, why `allow_blank`?

gabebw: Specifically: Is this to allow gabe -> edward (with edward pointing nowhere) instead of gabe -> edward -> gabe? I think that breaks the ring, right?

(Maybe there's another reason and I'm totally misreading this.)

edwardloveall: Oh, I think this was a concession to adding the first site. How can you have a next when there is no next to go to? But I think we can have next be itself. That should validate?

gabebw: Oh right, the first site. Yes, I think that'll validate.

app/models/redirection.rb
edwardloveall: It's not in previous (just removed it), but it is in next because `next` is a ruby keyword

edwardloveall:
@gabebw @adarsh okay give this whole thing a look-see and let me know if I'm good to :shipit:

app/models/redirection.rb
gabebw: Ah, yeah.

spec/controllers/redirections_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe RedirectionsController do
+  describe "GET :next" do
+    it "redirects to the next site" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      get :next, slug: "gabe"
```
gabebw: `gabe.slug`?

spec/controllers/redirections_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+  end
+
+  describe "GET :previous" do
+    it "redirects to the previous site" do
+      gabe = create(:redirection, :gabe)
+      edward = create(:redirection, :edward, next: gabe)
+      gabe.update(next: edward)
+
+      get :previous, slug: "edward"
```
gabebw: `slug: edward.slug`?

gabebw:
Minor comments, overall looks great. :100: :tada: :fireworks: :sparkling_heart:

PR #2, opened 2015-11-10
Rubocop

edwardloveall:
I code more better with this


gabebw:
Does `Enabled: false` mean that it checks for the opposite? Or can those cops be removed without changing the effect of this file?

(I don't use Rubocop.)

edwardloveall:
Not sure, TBH. I copied this directly from hound's ruby.yml

gabebw:
OK merge it

PR #5, opened 2015-11-11
Remove foreign key constraint

gabebw:
The foreign key constraint on the `next_id` column meant that we can't insert the first record, because any ID we put on next_id won't be a valid foreign key (since there's nothing in the table).

This migration removes that foreign key constraint so that we can insert the first few records and kickstart the ring.

All specs are passing.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -1,13 +1,13 @@
 FactoryGirl.define do
   factory :redirection do
-    trait :gabe do
-      slug "gabe"
-      url "http://gabebw.com"
-    end
+    sequence(:slug) { |n| "slug#{n}" }
+    sequence(:url) { |n| "http://example#{n}.com" }
+    next_id -1
 
-    trait :edward do
-      slug "edward"
-      url "http://edwardloveall.com"
+    after(:create) do |redirection|
+      if redirection.next_id == 0
+        Ring.new(redirection).link
+      end
     end
   end
 end
--- a/spec/models/ring_spec.rb
+++ b/spec/models/ring_spec.rb
@@ -3,29 +3,15 @@ require "rails_helper"
 RSpec.describe Ring do
   describe "#link" do
     it "adds a new redirection while preserving the ring" do
-      redirection = Redirection.create!(
-        url: "http://example.com/1",
-        slug: "one",
-        next_id: 0,
-      )
-      other_redirection = Redirection.create!(
-        url: "http://example.com/2",
-        slug: "two",
-        next: redirection,
-      )
-      redirection.update!(next_id: other_redirection.id)
-      new_redirection = Redirection.new(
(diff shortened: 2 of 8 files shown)


spec/factories.rb
```diff
@@ -1,5 +1,20 @@
 FactoryGirl.define do
   factory :redirection do
+    sequence(:slug) { |n| "slug#{n}" }
+    sequence(:url) { |n| "http://example#{n}.com" }
+    next_id 0
+
+    after(:create) do |redirection|
+      if redirection.next_id != 0
```
gabebw: This is a huge hack: if the ID is 0 (and thus obviously fake), set `next_id` to refer to a real Redirection.

edwardloveall: I still feel like we could use seeds, no? http://stackoverflow.com/a/19930700/638966

Then we can have seeds that we'll use for production anyway.

gabebw: Oh gosh that's way better. One sec.

edwardloveall: From offline conversation:

Gabe: Even if we use seeds, we still can't `create(:redirection)` without errors
Edward: What if we make the default `next` equal to `Redirection.last` since the seeds will have a couple of redirections already in place

gabebw: > What if we make the default next equal to Redirection.last since the seeds will have a couple of redirections already in place

I think that runs into the uniqueness constraint?

spec/support/database_cleaner.rb
```diff
@@ -1,6 +1,9 @@
 RSpec.configure do |config|
   config.before(:suite) do
-    DatabaseCleaner.clean_with(:deletion)
+    DatabaseCleaner.strategy = :transaction
+    # DatabaseCleaner.clean_with(:deletion)
```
edwardloveall: Comment?

gabebw: Good catch

edwardloveall:
K I think it looks good.

spec/models/ring_spec.rb
```diff
@@ -3,29 +3,15 @@
...
+      new_redirection = build(:redirection)
 
       ring = Ring.new(new_redirection)
       ring.link
 
-      [redirection, other_redirection].each(&:reload)
-      expect(redirection.next).to eq other_redirection
-      expect(other_redirection.next).to eq new_redirection
-      expect(new_redirection.next).to eq redirection
+      redirections = Redirection.order(:id).all
```
edwardloveall: Don't we want to order by date not id?

gabebw: It doesn't _really_ matter but yes we do

gabebw: Fixed

edwardloveall:
merge it

PR #6, opened 2015-11-12
Easily add another link to the ring

gabebw:
The `Ring` class takes care of breaking links, adding another redirection, and re-forming links to take into account the new entry.

It runs in a transaction so that if part of `Ring#link` fails, we're not left with junk data.

This should help with #5.

### DIFF
--- a/app/models/ring.rb
+++ b/app/models/ring.rb
@@ -1,0 +1,19 @@
+class Ring
+  def initialize(redirection_to_link)
+    @oldest_redirection = Redirection.order(created_at: :desc).last
+    @newest_redirection = Redirection.order(created_at: :desc).first
+    @redirection_to_link = redirection_to_link
+  end
+
+  def link
+    Redirection.transaction do
+      redirection_to_link.update!(next_id: 0)
+      newest_redirection.update!(next: redirection_to_link)
+      redirection_to_link.update!(next: oldest_redirection)
+    end
+  end
+
+  private
+
+  attr_reader :oldest_redirection, :newest_redirection, :redirection_to_link
+end
--- a/spec/models/ring_spec.rb
+++ b/spec/models/ring_spec.rb
@@ -1,0 +1,31 @@
+require "rails_helper"
+
+RSpec.describe Ring do
+  describe "#link" do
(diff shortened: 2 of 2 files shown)


app/models/ring.rb
```diff
@@ -0,0 +1,19 @@
+class Ring
+  def initialize(redirections, redirection_to_link)
+    @oldest_redirection = redirections.order(created_at: :desc).last
```
gabebw: I'll change the `order` to a scope named `newest_first

app/models/ring.rb
```diff
@@ -0,0 +1,19 @@
+class Ring
+  def initialize(redirections, redirection_to_link)
```
edwardloveall: Why pass in redirections? Do we not always want `Redirection.all`?

gabebw: > Why pass in redirections?

I over-architected for some dependency-injection future that doesn't exist now and is unlikely to in the future.

spec/models/ring_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+RSpec.describe Ring do
+  describe "#add" do
```
edwardloveall: I think you meant `#link`

gabebw: 4a9189776005b653631f7f6c1556250e79fbdead

edwardloveall:
Nice. Looks good to me.

app/models/ring.rb
```diff
@@ -0,0 +1,19 @@
+class Ring
+  def initialize(redirection_to_link)
+    @oldest_redirection = Redirection.order(created_at: :desc).last
```
adarsh: Random question: Why do this as an instance variable assignment in the constructor vs. a private method?

edwardloveall: I think the goal was to clean this up a little later. This was a really fast fix for some circular dependency problems we were having. Wouldn't be surprised if this class gets totally refactored. Personally, I'd like to see `Ring.link(...)` that calls `new(...).link` as a convenience method.

PR #8, opened 2015-11-29
Always break the ring at the first element

gabebw:
Previously we were doing some, uh, interesting stuff around newest/oldest Redirections, which assumed that the newest redirection links to the oldest one.  That's not necessarily true.

Now we:

* Grab the first Redirection (`@from_redirection`)
* Grab the Redirection that that Redirection points to (`@to_redirection`)
* Break the link between them
* Insert a new redirection between them

This ensures that we're breaking a link that definitely exists, instead of breaking a link that we assume (based on creation time) exists.

This commit also refactors the tests to use the existing gabe/edward Redirections that are created via DB seeds.

### DIFF
--- a/app/models/ring.rb
+++ b/app/models/ring.rb
@@ -1,19 +1,18 @@
 class Ring
   def initialize(redirection_to_link)
-    @oldest_redirection = Redirection.order(created_at: :desc).last
-    @newest_redirection = Redirection.order(created_at: :desc).first
     @redirection_to_link = redirection_to_link
+    @from_redirection = Redirection.first
+    @to_redirection = from_redirection.next
   end
 
   def link
     Redirection.transaction do
-      redirection_to_link.update!(next_id: 0)
-      newest_redirection.update!(next: redirection_to_link)
-      redirection_to_link.update!(next: oldest_redirection)
+      from_redirection.update!(next: redirection_to_link)
+      redirection_to_link.update!(next: to_redirection)
     end
   end
 
   private
 
-  attr_reader :oldest_redirection, :newest_redirection, :redirection_to_link
+  attr_reader :from_redirection, :to_redirection, :redirection_to_link
 end
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -3,25 +3,23 @@ require "rails_helper"
 RSpec.describe RedirectionsController do
   describe "GET :next" do
     it "redirects to the next site" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      gabe = Redirection.find_by!(slug: "gabe")
+      next_redirection = gabe.next
 
       get :next, slug: gabe.slug
 
-      expect(response).to redirect_to(edward.url)
+      expect(response).to redirect_to(next_redirection.url)
     end
   end
 
   describe "GET :previous" do
     it "redirects to the previous site" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      gabe = Redirection.find_by!(slug: "gabe")
+      previous = Redirection.find_by!(next: gabe)
 
-      get :previous, slug: edward.slug
+      get :previous, slug: gabe.slug
 
-      expect(response).to redirect_to(gabe.url)
+      expect(response).to redirect_to(previous.url)
     end
   end
 end
--- a/spec/models/redirection_spec.rb
+++ b/spec/models/redirection_spec.rb
@@ -23,21 +23,22 @@ RSpec.describe Redirection do
 
   describe "#next_url" do
     it "returns the url of the next referenced redirection" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      first = Redirection.first
+      redirection = build(:redirection)
+      Ring.new(redirection).link
 
-      expect(gabe.next_url).to eq(edward.url)
+      first.reload
+      expect(first.next_url).to eq(redirection.url)
     end
   end
 
   describe "#previous_url" do
     it "returns the url of the previous referenced redirection" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      first = Redirection.first
+      redirection = build(:redirection)
+      Ring.new(redirection).link
 
-      expect(gabe.previous_url).to eq(edward.url)
+      expect(first.previous_url).to eq("http://edwardloveall.com")
     end
   end
 end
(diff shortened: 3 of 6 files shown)


spec/controllers/redirections_controller_spec.rb
```diff
@@ -3,25 +3,23 @@
...
     end
   end
 
   describe "GET :previous" do
     it "redirects to the previous site" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      gabe = Redirection.find_by!(slug: "gabe")
+      previous = Redirection.find_by!(next: gabe)
```
gabebw: This is more explicit now.

app/models/ring.rb
```diff
@@ -1,19 +1,18 @@
 class Ring
   def initialize(redirection_to_link)
-    @oldest_redirection = Redirection.order(created_at: :desc).last
-    @newest_redirection = Redirection.order(created_at: :desc).first
     @redirection_to_link = redirection_to_link
+    @from_redirection = Redirection.first
+    @to_redirection = from_redirection.next
```
adarsh: What about setting these as methods?

I'm not sure if someone told me it was a good idea, but I'm only in the habit of doing param-to-ivar work in the constructor and putting everything else in a method. Interested in your feedback.

spec/controllers/redirections_controller_spec.rb
```diff
@@ -3,25 +3,23 @@
 RSpec.describe RedirectionsController do
   describe "GET :next" do
     it "redirects to the next site" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      gabe = Redirection.find_by!(slug: "gabe")
```
adarsh: Wait, do the seeds get created before each spec run or as part of `rake db:test:prepare`? I'm not used to accessing AR objects created outside of the spec block.

spec/models/redirection_spec.rb
```diff
@@ -33,11 +32,10 @@
...
   describe "#previous_url" do
     it "returns the url of the previous referenced redirection" do
-      gabe = create(:redirection)
-      edward = create(:redirection, next: gabe)
-      gabe.update(next: edward)
+      gabe = Redirection.find_by!(slug: "gabe")
+      edward = Redirection.find_by!(slug: "edward")
 
-      expect(gabe.previous_url).to eq(edward.url)
+      expect(edward.previous_url).to eq(gabe.url)
```
adarsh: All looks good, but may be worth considering a spec where the ring has more than two members, just in case.

spec/models/ring_spec.rb
```diff
@@ -4,14 +4,16 @@
   describe "#link" do
     it "adds a new redirection while preserving the ring" do
       new_redirection = build(:redirection)
+      redirections = Redirection.order(created_at: :asc)
```
adarsh: This variable seems to be unused but I may be misreading.

spec/models/ring_spec.rb
```diff
@@ -4,14 +4,16 @@
   describe "#link" do
     it "adds a new redirection while preserving the ring" do
       new_redirection = build(:redirection)
```
adarsh: Why `build` here instead of `create`? Is it to do with callbacks for integrating into the ring?

adarsh:
LGTM, although this one is new to me: 

> This commit also refactors the tests to use the existing gabe/edward Redirections that are created via DB seeds.

Feels like fixtures (that is not a troll, I swear).

spec/controllers/redirections_controller_spec.rb
edwardloveall: [Yes](https://github.com/gabebw/hotline-webring/blob/master/spec%2Fsupport%2Fdatabase_cleaner.rb#L5). It hides some logic, but there are a lot of things we can assume if there are two redirections in the database, which is the case.

spec/models/ring_spec.rb
edwardloveall: I think it's generally better to use build where you can because the you don't have to hit the DB as much.

edwardloveall:
How is this way of building the ring better? Honestly wondering. It certainly creates a weird chain after it's done:

```
first
fifth
fourth
third
second
```

¯\_(ツ)_/¯

spec/models/ring_spec.rb
edwardloveall: I think you're right

edwardloveall:
After thinking about it, I think I was just getting confused in the Ring code. From the perspective of a visitor, the order is no different than:

```
fifth
fourth
third
second
first
```

which is fine.

gabebw:
> Feels like fixtures (that is not a troll, I swear).

It's totally fixtures, albeit fixtures that are also created when you run bin/setup (so not just in the test environment). For the Ring to work, we need SOME seed data, and I thought this was the best way to do it.

app/models/ring.rb
gabebw: > What about setting these as methods?

What do you mean? Can you provide example code?

spec/models/ring_spec.rb
gabebw: Yup, fixed: 6df711cc46610702ada76b9158e133c8785d8102

spec/models/ring_spec.rb
gabebw: It's because of callbacks for integrating into the ring, yes. It's not really because of hitting the DB.

spec/models/redirection_spec.rb
gabebw: Done: e18accfaa7c8517dd28942323c819daef2023730

gabebw:
Ready for re-review.

spec/models/redirection_spec.rb
gabebw: Oh, oops, you mean in here.

gabebw: How's this? 7322690c786b87124ff57e63f4042ef98f155d6f

app/models/ring.rb
edwardloveall: I think he means:

```ruby
private

def from_redirection
  @from_redirection ||= Redirection.first
end

def to_redirection
  @to_redirection ||= from_redirection.next
end
```

gabebw: Sure, I like not doing DB access in the constructor.

spec/models/ring_spec.rb
edwardloveall: My b.

app/models/ring.rb
gabebw: Actually, that code snippet makes everything break, probably because it's grabbing `Redirection.first.next` while we're fiddling with the ring. We need to grab `Redirection.first.next` before doing anything fancy, i.e. in the constructor.

PR #9, opened 2015-12-02
Auto-create redirections

gabebw:
When we receive a request for `/new-slug/next` or `/new-slug/previous`, we:

* check if there's an entry for `new-slug`
* if there isn't, we add it to the ring
* redirect to the next/previous URL for `new-slug`

### DIFF
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -12,6 +12,58 @@ RSpec.describe RedirectionsController do
     end
   end
 
+  [:next, :previous].each do |action|
+    describe "GET #{action}" do
+      it "creates a Redirection if it doesn't exist" do
+        new_slug = "new"
+        url = "http://example.com"
+        first_redirection = Redirection.first
+        old_next = first_redirection.next
+
+        request.env["HTTP_REFERER"] = url
+        get action, slug: new_slug
+
+        new_redirection = Redirection.find_by!(slug: new_slug)
+        first_redirection.reload
+        expect(first_redirection.next.slug).to eq new_slug
+        expect(new_redirection.url).to eq url
+        expect(new_redirection.next).to eq old_next
+      end
+
+      it "uses the referrer's hostname, including subdomain" do
+        new_slug = "new"
+        hostname = "http://cool.example.com"
+
+        request.env["HTTP_REFERER"] = "#{hostname}/something/else"
+        get action, slug: new_slug
+
+        expect(Redirection.find_by!(slug: new_slug).url).to eq hostname
+      end
+
+      context "when there is no referrer" do
+        it "does not create a Redirection" do
+          new_slug = "new"
+
+          get action, slug: new_slug
+
+          expect(Redirection.exists?(slug: new_slug)).to be false
+        end
+
+        it "redirects to the first redirection's next/previous URL" do
+          new_slug = "new"
+
+          get action, slug: new_slug
+
+          if action == :next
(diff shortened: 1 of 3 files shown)


spec/controllers/redirections_controller_spec.rb
```diff
@@ -10,6 +10,20 @@
...
+
+    it "creates a Redirection if it doesn't exist" do
+      new_slug = "new"
+      url = "http://example.com"
+      old_next_url = Redirection.first.next_url
+
+      request.env["HTTP_REFERER"] = url
+      get :next, slug: new_slug
+
+      expect(Redirection.first.next.slug).to eq new_slug
```
edwardloveall: Maybe put some of these `Redirection`s into variables? Some of the method chaining is hard to read.

spec/controllers/redirections_controller_spec.rb
```diff
@@ -10,6 +10,20 @@
 
       expect(response).to redirect_to(next_redirection.url)
     end
+
+    it "creates a Redirection if it doesn't exist" do
+      new_slug = "new"
+      url = "http://example.com"
+      old_next_url = Redirection.first.next_url
+
+      request.env["HTTP_REFERER"] = url
```
edwardloveall: Will the referrer be `http://example.com` or `http://example.com/foo/next`?

edwardloveall:
Sweet! You did this so I didn't have to.

edwardloveall:
Looks pretty good though. Glad we really ironed out all the Ring stuff.

spec/controllers/redirections_controller_spec.rb
gabebw: Fixed in 601fedbfb3557e889462ba33000073d847923bfc

gabebw:
@edwardloveall I remember we discussed using `/gabebw/next` instead of just `/next` so that `blog.gabebw.com` and `gabebw.com` can be part of the webring with the same slug.

But the first one of `blog.gabebw.com`/`gabebw.com` that someone clicks on will be the URL for the `gabebw` slug. I.e. if someone clicks on the `/gabebw/next` link from my blog, then the Redirection for `gabebw` will have a URL of `blog.gabebw.com`, and `gabebw.com` will never be linked from any other site in the webring.

We could instead have links like `/next`, without a slug, and use the full domain name of the referrer (`blog.gabebw.com`, `gabebw.com`) as the slug. This would put both sites in the webring, but does mean we get the "double dipping" you were worried about.

Thoughts?

spec/controllers/redirections_controller_spec.rb
gabebw: Better? It's still magicky: 8073db71e24d472a40a077f79c373ec6f6b33e4d

edwardloveall:
Here are some ideas that I just thought of:

A) Having some sort of moderation, presumably by us, that if people made some mistake they can fix it
B) Design/Write some pretty decent documentation where we warn people that the first site that is linked the site that will be in the ring
C) Have people use a DNS txt record to validate a domain
D) Have some weird removal url that people can use to delete and try again

B) has some pretty annoying customer frustration potential. I'd say A) even if it's just rails console or C) if we really want to do a lot of work

It also brings up in my head how do we remove people that don't have active webring links, but that is for another time.

gabebw:
> C) Have people use a DNS txt record to validate a domain

That is a very Us response (cf. `dig TXT reply.gabebw.com +short`).

adarsh:
Shipped is better than Perfect™.

PR #10, opened 2015-12-03
Extract a RedirectionCreation service

gabebw:
RedirectionCreation handles creating (or not creating!) a new Redirection based on a slug and referrer. It also links the new redirection into the ring.

Since we handle edge cases (like a blank referrer) at the RedirectionCreation unit test level, the controller specs can now be correspondingly simpler.

### DIFF
--- a/app/services/redirection_creation.rb
+++ b/app/services/redirection_creation.rb
@@ -1,0 +1,29 @@
+class RedirectionCreation
+  def self.perform(referrer, slug)
+    new(referrer, slug).perform
+  end
+
+  def initialize(referrer, slug)
+    @referrer = referrer
+    @slug = slug
+  end
+
+  def perform
+    redirection = Redirection.new(slug: slug)
+
+    if referrer.present?
+      redirection.url = referrer_hostname
+      Ring.new(redirection).link
+      redirection
+    end
+  end
+
+  private
+
+  attr_reader :referrer, :slug
+
+  def referrer_hostname
+    referrer_uri = URI.parse(referrer)
+    "#{referrer_uri.scheme}://#{referrer_uri.hostname}"
+  end
+end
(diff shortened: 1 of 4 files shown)


app/services/redirection_creation.rb
```diff
@@ -0,0 +1,25 @@
...
+      redirection
+    end
+  end
+
+  private
+
+  attr_reader :referrer, :slug
+
+  def referrer_hostname
+    referrer_uri = URI.parse(referrer)
```
edwardloveall: I only knew about `URI(referrer)`. `.parse` is probably slightly more clear anyway.

app/services/redirection_creation.rb
```diff
@@ -0,0 +1,25 @@
+class RedirectionCreation
+  def initialize(referrer, slug)
+    @referrer = referrer
+    @slug = slug
+  end
+
+  def perform
```
edwardloveall: I'd also recommend a class method to go along side this:

```ruby
def self.perform(referrer, slug)
  new(referrer, slug).perform
end
```

because it makes future testing (stubbing in particular) much easier. I.e. you don't have to stub `new` to return a double then stub out `perform` on the double. Also, if for whatever reason we use background jobs, I hear having those `perform` methods can make that easier too.

edwardloveall:
Looks pretty good to me :red_circle: :black_circle: :white_circle: :large_blue_circle:

app/services/redirection_creation.rb
gabebw: Done!

edwardloveall:
Yeah it's lookin good in this neighborhood

app/services/redirection_creation.rb
```diff
@@ -0,0 +1,29 @@
+class RedirectionCreation
+  def self.perform(referrer, slug)
+    new(referrer, slug).perform
```
adarsh: Beauty!

PR #13, opened 2015-12-06
Repeat the header text, like the Drake video

gabebw:
The video repeats "HOTLINE BLING" 6 times. (I counted.)

### DIFF
--- a/app/views/homes/show.html.erb
+++ b/app/views/homes/show.html.erb
@@ -1,4 +1,8 @@
-<h1><%= title %></h1>
+<header>
+  <% 6.times do %>
+    <h2><%= title %></h2>
+  <% end %>
+</header>
 
 <p>Wanna join the coolest webring? Here's how.</p>
 
@@ -8,7 +12,7 @@
   <li>You're done! When someone clicks on one of your links, you'll be automatically added to the webring.</li>
 </ol>
 
-<h2>FAQ</h2>
+<h3>FAQ</h3>
 
 <p>
   Q: Can I control who I link to?</br>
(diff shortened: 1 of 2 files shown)


app/views/homes/show.html.erb
```diff
@@ -1,18 +1,22 @@
-<h1><%= title %></h1>
+<% 6.times do %>
+  <h1><%= title %></h1>
```
gabebw: Ideally we'd have one h1 per page. But in this case, every h1 has the same text. So I think it's OK?

I welcome opinions.

gabebw:
Ideally we'd have one h1 per page. But in this case, every h1 has the same text. So I think it's OK?

I welcome opinions.

edwardloveall:
I suppose that that point it's not really fulfilling its role as a heading and could be a bunch of spans instead. How accessible is this webring?

gabebw:
> How accessible is this webring?

Uh, pretty accessible I think. It's mostly a bunch of URLs.

> I suppose that that point it's not really fulfilling its role as a heading and could be a bunch of spans instead

Hmm, spans feel odd, because it is _a_ header, just not _the_ header. What about h2s?

edwardloveall:
`h2`s are fine too. I was thinking spans they don't denote any sort of heading. I think there's a way to hack extra text with css...

I'm hilariously imagining a screenreader reading "Hotline webring hotlinewebringhotlinewebringhotlinewebringhotlinewebringhotlinewebring"

gabebw:
Done, with h2s: 46b239864bc85953c622bffdbc816cfc56b2619f

edwardloveall:
:shipit:

PR #14, opened 2015-12-09
Homepage links & slugs

edwardloveall:
This adds the contents of the current webring on the homepage so people A) have a place to point and say "Hey that's me!" and for potential users to see what slugs already exist.

It looks like this:
screen shot 2015-12-08 at 6 08 05 pm

### DIFF
--- a/app/assets/stylesheets/base/_variables.scss
+++ b/app/assets/stylesheets/base/_variables.scss
@@ -21,13 +21,14 @@ $dark-gray: #333;
 $medium-gray: #999;
 $light-gray: #ddd;
 $hotline-bling-pink: #FA9ABA;
+$dark-pink: darken($hotline-bling-pink, 40%);
 
 // Font Colors
 $base-font-color: $dark-gray;
-$action-color: $blue;
+$action-color: black;
 
 // Border
-$base-border-color: $light-gray;
+$base-border-color: $dark-pink;
 $base-border: 1px solid $base-border-color;
 
 // Background Colors
--- a/app/controllers/homes_controller.rb
+++ b/app/controllers/homes_controller.rb
@@ -1,4 +1,5 @@
 class HomesController < ApplicationController
   def show
+    @redirections ||= Redirection.in_ring_order
   end
 end
--- a/app/views/redirections/_redirection.html.erb
+++ b/app/views/redirections/_redirection.html.erb
@@ -1,0 +1,4 @@
+<tr>
+  <td><%= redirection.slug %></td>
+  <td><%= link_to(redirection.url) %></td>
+</tr>
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -6,7 +6,9 @@ if Rails.env.development? || Rails.env.test?
     task prime: "db:setup" do
       include FactoryGirl::Syntax::Methods
 
-      # create(:user, email: "user@example.com", password: "password")
+      create(:redirection, slug: "adarsh", url: "http://adarsh.io")
+      create(:redirection, slug: "thorncp", url: "http://thorn.co")
+      create(:redirection, slug: "foxed", url: "http://alexfox.me")
     end
   end
 end
--- a/spec/models/redirection_spec.rb
+++ b/spec/models/redirection_spec.rb
@@ -21,6 +21,17 @@ RSpec.describe Redirection do
     end
   end
 
+  describe ".in_ring_order" do
+    it "returns redirections in the order they appear next in the ring" do
+      create(:redirection)
+      first = Redirection.order(id: :asc).first
+      next_one = first.next
+      final = next_one.next
+
+      expect(Redirection.in_ring_order).to match_array([first, next_one, final])
+    end
+  end
+
   describe "#next_url" do
     it "returns the url of the next referenced redirection" do
       first = Redirection.first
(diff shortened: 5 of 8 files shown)


app/assets/stylesheets/base/_variables.scss
```diff
@@ -21,13 +21,14 @@ $dark-gray: #333;
 $medium-gray: #999;
 $light-gray: #ddd;
 $hotline-bling-pink: #FA9ABA;
+$dark-pink: darken($hotline-bling-pink, 40%);
```
gabebw: Could we darken it even more? The contrast is a little low.

app/views/redirections/_redirection.html.erb
```diff
@@ -0,0 +1,6 @@
+<tr>
+  <%= content_tag(:td, redirection.slug) %>
```
gabebw: Why not `<td><%= redirection.slug %></td>`?

spec/models/redirection_spec.rb
```diff
@@ -21,6 +21,17 @@
     end
   end
 
+  describe ".in_ring_order" do
+    it "returns redirections in the order they appear next in the ring" do
+      create(:redirection)
+      first = Redirection.order(id: :asc).first
```
gabebw: I think `.first` orders it this way by default (as of Rails 4).

gabebw: http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-first

app/assets/stylesheets/base/_variables.scss
```diff
@@ -21,13 +21,14 @@ $dark-gray: #333;
...
 
 // Font Colors
 $base-font-color: $dark-gray;
-$action-color: $blue;
+$action-color: $dark-pink;
 
 // Border
-$base-border-color: $light-gray;
+$base-border-color: $dark-pink;
```
gabebw: Nice.

app/controllers/homes_controller.rb
```diff
@@ -1,4 +1,11 @@
 class HomesController < ApplicationController
   def show
   end
+
+  private
+
+  def redirections
+    @_redirections ||= Redirection.in_ring_order
+  end
+  helper_method :redirections
```
gabebw: Love the `helper_method` next to it.

edwardloveall: Sure! I did it this way because Sean Griffin tries to avoid ivars in controller methods. I can't remember the exact reasoning but it made sense in my head when I heard it.

app/views/redirections/_redirection.html.erb
edwardloveall: I've kind of gotten into the habit of using the `content_tag` and `content_tag_for` methods. I find them somewhat cleaner than switching back and forth between html and erb. I dunno, it's something I'm trying out to see if it's any better. Do you like the other way better or are you just curious?

spec/models/redirection_spec.rb
edwardloveall: True, but I think it shows some intent here. I want the record with the lowest numbered ID rather than "whatever the first one by some metric" is.

gabebw: Excellent point, this line is fine then.

app/views/redirections/_redirection.html.erb
gabebw: > Do you like the other way better or are you just curious?

When I see `content_tag`, I assume something fancy is happening. In this case, nothing fancy is happening, so it confused me for a sec.

app/assets/stylesheets/base/_variables.scss
gabebw: Maybe we could use gray for links, or just underline links and make them black?

lib/tasks/dev.rake
```diff
@@ -6,7 +6,9 @@ if Rails.env.development? || Rails.env.test?
     task prime: "db:setup" do
       include FactoryGirl::Syntax::Methods
 
-      # create(:user, email: "user@example.com", password: "password")
+      create(:redirection, slug: "adarsh", url: "http://adarsh.io")
```
gabebw: Should these be linked?

app/assets/stylesheets/base/_variables.scss
edwardloveall: Gray on pink doesn't look gray-t.

Black works fine. I can change them to that.

edwardloveall: Fixed in 22b324d

app/views/redirections/_redirection.html.erb
edwardloveall: Fixed in 6aa2d3f

lib/tasks/dev.rake
edwardloveall: They get linked in the factory: https://github.com/gabebw/hotline-webring/blob/master/spec%2Ffactories.rb#L9

app/controllers/homes_controller.rb
edwardloveall: Fixed in 1f8c88b

edwardloveall:
Ready for re-review

gabebw:
Great commit messages. :+1:

PR #16, opened 2015-12-09
Fix regression on homepage

edwardloveall:
In a previous (now squashed) commit, I had a helper method for
`redirections` on the homepage to avoid an ivar. However, I changed
back to an ivar in the `HomesController` but not in the view. The app
crashed on the homepage.

This is now tested for.

### DIFF
--- a/spec/features/user_visits_homepage_spec.rb
+++ b/spec/features/user_visits_homepage_spec.rb
@@ -4,8 +4,6 @@ RSpec.feature "User visits homepage" do
   scenario "sees list of redirections" do
     visit root_path
 
-    expect(page.html).to have_css("table.redirections")
-
     within("table.redirections") do
       expect(page).to have_content("gabebw")
       expect(page).to have_content("http://gabebw.com")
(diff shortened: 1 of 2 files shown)


spec/features/user_visits_homepage_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+
+RSpec.feature "User visits homepage" do
+  scenario "sees list of redirections" do
+    visit root_path
+
+    expect(page.html).to have_css("table.redirections")
```
gabebw: This line feels more like it belongs in a view spec -- it doesn't actually break if you remove this, right? I think the `within` will break if this isn't present, too.

What do you think about removing this line and keeping the rest of the spec as-is

edwardloveall: Oh yeah... I forgot view specs were a thing. Okay, will remove

edwardloveall: Removed in 71320fd

gabebw:
:+1:

PR #22, opened 2015-12-16
Make header styling consistent everywhere

gabebw:
Instead of only styling headers for the landing page, style them across the entire site.

PR #23, opened 2015-12-16
Ignore people arriving from localhost

gabebw:
Please check out #22 first.

When people click links on their site running on `http://localhost:3000`, we were adding `http://localhost` to the webring.

Now we don't add localhost requests to the webring, and we provide a helpful explanation of what's going on.

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -1,4 +1,6 @@
 class RedirectionsController < ApplicationController
+  before_action :ensure_referrer_is_not_localhost
+
   def next
     redirection = find_or_create_redirection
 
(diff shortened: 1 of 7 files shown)


spec/controllers/redirections_controller_spec.rb
```diff
@@ -28,6 +28,14 @@
         expect(first_redirection.reload.next).to eq new_redirection
       end
 
+      it "ignores requests from localhost" do
+        request.env["HTTP_REFERER"] = "http://localhost:3000"
+
+        get action, slug: "whatever"
```
adarsh: whatever

adarsh:
:+1:

app/controllers/redirections_controller.rb
```diff
@@ -1,4 +1,6 @@
 class RedirectionsController < ApplicationController
+  before_action :ensure_referrer_is_not_localhost
```
edwardloveall: An alternative to the before action would be to:

* Make a validator that made sure the host wasn't localhost
* Raise a validation error if it was
* `rescue_from` that error in the app controller and redirect to localhost page

I've used this pattern in APIs to great effect, and mostly avoided before_actinos. Do you have a preference?

gabebw:
Agh, closing #22 closed this and made it un-reopenable. GitHub!

gabebw:
Reopened as #24.

PR #24, opened 2015-12-16
Ignore people arriving from localhost

gabebw:
When people click links on their site running on `http://localhost:3000`, we were adding `http://localhost` to the webring.

Now we don't add localhost requests to the webring, and we provide a helpful explanation of what's going on.

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -1,4 +1,6 @@
 class RedirectionsController < ApplicationController
+  before_action :ensure_referrer_is_not_localhost
+
   def next
     redirection = find_or_create_redirection
 
@@ -21,5 +23,11 @@ class RedirectionsController < ApplicationController
 
   def referrer
     request.env["HTTP_REFERER"]
+  end
+
+  def ensure_referrer_is_not_localhost
+    if referrer.present? && URI.parse(referrer).host == "localhost"
+      redirect_to page_path("localhost")
+    end
   end
 end
--- a/app/assets/stylesheets/_landing-page.scss
+++ b/app/assets/stylesheets/_landing-page.scss
@@ -11,11 +11,6 @@
     }
   }
 
-  .explanation {
-    @include shift(2);
-    @include span-columns(8);
-  }
-
   iframe {
     margin-top: 2em;
   }
--- a/app/assets/stylesheets/_mobile.scss
+++ b/app/assets/stylesheets/_mobile.scss
@@ -16,7 +16,7 @@
   }
 
   .homes-show {
-    .explanation {
+    .text {
       @include span-columns(3);
     }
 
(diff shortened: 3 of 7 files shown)


app/controllers/redirections_controller.rb
```diff
@@ -1,4 +1,6 @@
 class RedirectionsController < ApplicationController
+  before_action :ensure_referrer_is_not_localhost
```
gabebw: Edward's comment from #23, reproduced below:

--------------------------------------------------------------------


An alternative to the before action would be to:

* Make a validator that made sure the host wasn't localhost
* Raise a validation error if it was
* `rescue_from` that error in the app controller and redirect to localhost page

I've used this pattern in APIs to great effect, and mostly avoided before_actinos. Do you have a preference?

gabebw: @edwardloveall, do you mean something like this code?

```ruby
class RedirectionsController
  rescue_from ValidationError, with: :redirect_to_localhost

  def next
    Validator.new(referrer).validate! # ValidationError comes from here

    redirection = find_or_create_redirection

    redirect_to redirection.next_url
  end

  private

  def redirect_to_localhost
    # ...
  end
end
```

I find `rescue_from` hard to reason about:

* When reading the `next` method, it's not clear that `Validator.new.validate!` can trigger a jump to a different method, because there's no `rescue` in that method.
* When reading the `rescue_from`, I personally can't tell where the `ValidationError` might come from (it could even be triggered by a method in `ApplicationController`). I must reason about the whole class to figure out what triggers the error and where it jumps from.
* `rescue_from` is using exceptions as control flow. Here's [a pretty good summary][so] of some arguments against exceptions as control flow. And [here's another][c2].

[so]: http://stackoverflow.com/a/729404
[c2]: http://c2.com/cgi/wiki?DontUseExceptionsForFlowControl

I don't have the same problems with `before_action`:

* `before_action`s are executed at a well-defined time (before each request)
* They cannot control the execution flow once a request "gets past" them -- if a controller method like `next` is being executed, the `before_action` can have no further effect.

That said, I'd be open to a plain `if` statement instead of the `before_action`:

```ruby
def next
  if referrer_is_localhost?
    redirect_to page_path("localhost")
  else
    # current implementation
  end
end
```

edwardloveall: Good points. I've used rescue_from to great effect before, but this is probably not the place for it.

What about:

```ruby
class Redirection < ActiveRecord::Base
  validates :url, format: { with: -> /regex for no localhost/ }
end
```

and then redirect if it doesn't validate with that format?

gabebw: What would that look like in the controller?

edwardloveall: You know, in my head I was thinking something like `if redirection.save` and then handling it there, but that's not what we have. Let's just go with this. We can always change it if we need to.

edwardloveall:
:shipit:

gabebw:
Thanks for the thorough review!

PR #25, opened 2015-12-17
Mobile-ify the layout

gabebw:
On mobile screens:

* Hide the giant video
* Use a smaller header text so each "HOTLINE WEBRING" header is on its own line instead of breaking over multiple lines

I tested it with Safari's (extremely handy) Responsive Design Mode, and the site looks good on everything from an iPhone 4S to an iPad Pro.

### DIFF
--- a/app/assets/stylesheets/_mobile.scss
+++ b/app/assets/stylesheets/_mobile.scss
@@ -1,0 +1,27 @@
+@include media($large-screen-down) {
+  h2 {
+    font-size: 3.5em;
+  }
+
+  .homes-show {
+    iframe {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/_mobile.scss
```diff
@@ -0,0 +1,27 @@
+@include media($large-screen-down) {
+  h2 {
```
gabebw: This isn't in `.homes-show` so that the `h2` in #24 will be smaller too.

edwardloveall:
Looks good!

PR #26, opened 2015-12-17
Remove delayed_job

gabebw:
* We don't have background jobs and are unlikely to add them
* Now Foreman no longer has to boot a worker process locally


edwardloveall:
:+1:

PR #28, opened 2015-12-21
Add RSpec formatter for Circle CI

gabebw:
With this formatter, Circle should pretty-print our failures. We need a custom spec command because otherwise Circle only runs RSpec, while `bundle exec rake` runs `rake bundle:audit` as well.

https://circleci.com/docs/test-metadata#automatic-test-metadata-collection

I also changed the specs so we can see how failures look now. I'll remove the failures before merging.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,7 @@ group :test do
   gem "database_cleaner"
   gem "formulaic"
   gem "launchy"
+  gem "rspec_junit_formatter"
   gem "shoulda-matchers"
   gem "simplecov", require: false
   gem "timecop"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -206,6 +206,9 @@ GEM
       rspec-mocks (~> 3.3.0)
       rspec-support (~> 3.3.0)
     rspec-support (3.3.0)
+    rspec_junit_formatter (0.2.3)
+      builder (< 4)
+      rspec-core (>= 2, < 4, != 2.12.0)
     safe_yaml (1.0.4)
     sass (3.4.19)
     sass-rails (5.0.4)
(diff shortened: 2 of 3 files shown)


edwardloveall:
circle died

gabebw:
Do you mean the failures? I added intentional failures (aedb2b4). Or do you mean something else?

edwardloveall:
Oh okay. I've never done anything with circle before so this is all new.

edwardloveall:
I think it looks good then? I'm not sure why we would want an intentional failure after we've seen what one looks like...

gabebw:
>  I'm not sure why we would want an intentional failure after we've seen what one looks like

If you look at the failure from this branch: https://circleci.com/gh/gabebw/hotline-webring/116
vs the failure from older branches: https://circleci.com/gh/gabebw/hotline-webring/110

At the top of build 116 (the first one), Circle nicely formatted the failures for us and we don't have to scroll down to see them, like we do in build 110. I added (temporary) failures so we could see them nicely formatted, since otherwise we wouldn't have any failures because we run clean :leaves:.

edwardloveall:
Cool! Yeah that looks good, but I think I'm asking the wrong question.

I assume you added the formatter so that in the future, when tests fail, we have easy to read failure messages. This branch is has a failing test now so we can test that the formatter worked. If we merge it without changes, we'll have a failing test in our code. Do you plan to remove the failing test now that we know the formatter works?

gabebw:
> Do you plan to remove the failing test now that we know the formatter works?

Yep, from the PR description :smile: : 

> I also changed the specs so we can see how failures look now. I'll remove the failures before merging.

edwardloveall:
Oh okay I missed that. Thanks! Looks good to me!

edwardloveall:
This has been a weekly demonstration on miscommunication. Now back to our regularly scheduled webring.

PR #29, opened 2015-12-31
Ignore a wider variety of local requests

gabebw:
Previously we were only checking for referrers from `localhost`. Now we check for:

* `localhost` (like `http://localhost:3000`)
* `127.0.0.1` (like `http://127.0.0.1:3000`)
* `*.dev` (like `http://example.dev`)

Introduce a `HostValidator` model to cover the new logic.

### DIFF
--- a/app/models/host_validator.rb
+++ b/app/models/host_validator.rb
@@ -1,0 +1,32 @@
+class HostValidator
+  INVALID_HOSTS = %w(127.0.0.1 localhost)
+  INVALID_TLDS = %w(.dev)
+
+  def initialize(url)
+    @url = url
+  end
+
+  def invalid?
+    if url.present?
+      invalid_host? || invalid_tld?
+    else
+      false
+    end
+  end
+
+  private
+
+  attr_reader :url
+
+  def invalid_host?
+    INVALID_HOSTS.include?(host)
+  end
+
+  def invalid_tld?
+    INVALID_TLDS.include?(File.extname(host))
+  end
+
+  def host
+    URI.parse(url).host
+  end
+end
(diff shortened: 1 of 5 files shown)


app/views/pages/localhost.html.erb
```diff
@@ -7,13 +7,13 @@
...
-    definitely not what you want.
+    something like <code>http://localhost</code> as your entry in the webring,
+    which is definitely not what you want.
   </p>
   <p>
-    To keep localhost out of the webring, we ignore requests from localhost and
-    don't add them to the webring.  But don't worry! When you click on your real
-    site it'll work!
+    To keep local hosts out of the webring, we ignore requests from localhost
+    (and a few others, like <code>127.0.0.1</code>) and don't add them to the
```
edwardloveall: Maybe link to the constants in host_validator or the file it self? They may get out of date but at least it's a complete list.

edwardloveall: Rather the link's location might get out of date.

app/models/host_validator.rb
```diff
@@ -0,0 +1,30 @@
...
+  private
+
+  attr_reader :url
+
+  def invalid_host?
+    INVALID_HOSTS.include?(host)
+  end
+
+  def invalid_tld?
+    INVALID_TLDS.include?(File.extname(host))
```
edwardloveall: Clever.

app/views/pages/localhost.html.erb
gabebw: I actually intentionally left this vague -- we'll probably always block localhost, and this lets people know that we also block a few other similar hosts while letting us get away with not enumerating all of them.

app/models/host_validator.rb
gabebw: :blush:

app/views/pages/localhost.html.erb
edwardloveall: Okay!

edwardloveall:
circle dead, but otherwise looks good to me

PR #30, opened 2016-01-14
Atom feed

edwardloveall:
This is a relatively simple feed that shows "gabebw joined Hotline
Webring" with a link to the site that joined.

I added a new key to our en.yml file for authors because someone's
gotta author these amazing feed entries.

### DIFF
--- a/app/models/redirection.rb
+++ b/app/models/redirection.rb
@@ -7,6 +7,11 @@ class Redirection < ActiveRecord::Base
   validates :url, presence: true
 
   scope :in_ring_order, -> { order(next_id: :desc) }
+  scope :latest_first, -> { order(created_at: :desc) }
+
+  def self.most_recent
+    latest_first.first
+  end
 
   def next_url
     self.next.url
@@ -14,5 +19,11 @@ class Redirection < ActiveRecord::Base
 
   def previous_url
     previous.url
+  end
+
+  def tag_uri
+    date = created_at.iso8601
+
+    "tag:#{UNCHANGING_HOSTNAME},#{date}:#{url}"
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,4 +1,5 @@
 en:
+  authors: Gabe and Edward
   date:
     formats:
       default:
--- a/spec/models/redirection_spec.rb
+++ b/spec/models/redirection_spec.rb
@@ -32,6 +32,24 @@ RSpec.describe Redirection do
     end
   end
 
+  describe ".latest_first" do
+    it "returns the redirections with the most recently created first" do
+      redirection = create(:redirection)
+      edward = Redirection.find_by(slug: "edward")
+      gabe = Redirection.find_by(slug: "gabe")
+
+      expect(Redirection.latest_first).to eq([redirection, edward, gabe])
+    end
+  end
+
+  describe ".most_recent" do
+    it "returns the one, most recently created redirection" do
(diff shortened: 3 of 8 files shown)


app/models/redirection.rb
```diff
@@ -7,6 +7,8 @@ class Redirection < ActiveRecord::Base
   validates :url, presence: true
 
   scope :in_ring_order, -> { order(next_id: :desc) }
+  scope :latest_first, -> { order(created_at: :desc) }
+  scope :most_recent, -> { order(created_at: :desc).first }
```
gabebw: I'm not sure this is a scope -- it doesn't return a chainable `ActiveRecord::Relation`, it returns a specific record. How about:

```ruby
class Redirection
  def most_recent
    self.class.latest_first.first
  end
end
```

app/models/redirection.rb
```diff
@@ -15,4 +17,11 @@ def next_url
   def previous_url
     previous.url
   end
+
+  def tag_uri
+    unchanging_hostname = "hotlinewebring.club"
```
gabebw: The hostname is in `ENV["APPLICATION_HOST"]`. How about:

```ruby
hostname = ENV.fetch("APPLICATION_HOST")
```

Then it'll work on staging, too.

config/locales/en.yml
```diff
@@ -1,4 +1,5 @@
 en:
+  authors: Gabe and Edward
```
gabebw: it me

spec/models/redirection_spec.rb
```diff
@@ -32,6 +32,24 @@
     end
   end
 
+  describe ".latest_first" do
+    it "returns the redirections with the most recently created first" do
+      redirection = create(:redirection)
+      edward = Redirection.find_by(slug: 'edward')
```
gabebw: double quotes plz

spec/models/redirection_spec.rb
```diff
@@ -52,4 +70,15 @@
       expect(first.previous_url).to eq("http://edwardloveall.com")
     end
   end
+
+  describe "#tag_uri" do
+    it "returns a properly formatted id string" do
+      redirection = create(:redirection,
```
gabebw: Formatting:

```ruby
redirection = create(
  :redirection,
  created_at: DateTime.new(1999, 2),
   url: "http://example.com",
)
```

app/models/redirection.rb
edwardloveall: Sorry, I should have put this in the PR description instead of just the commit message, but this shouldn't ever change, even if the domain does:

>According to the [atom standard](http://tools.ietf.org/html/rfc4287#page-20) entries in an atom feed should never, ever change. They should also not be the same as the full URI (or really even be a URI) to limit confusion as to what they are used for. They are a unique string used for identification.

>`hotlinewebring.club` is hardcoded into for tag uri formula so that even if the app's domain changes, the unique, never changing ids will stay the same. Hopefully.

> Here is a blog post for reference: https://web.archive.org/web/20081225232147/http://diveintomark.org/archi
ves/2004/05/28/howto-atom-id

edwardloveall:
Ready for re-review

app/models/redirection.rb
gabebw: Oh interesting! How about making this a constant then?

edwardloveall: mmm yup! that'll do

gabebw:
:+1: squash & merge

PR #31, opened 2016-02-02
Force SSL in production (and staging)

gabebw:
Now that Cloudflare gives us SSL, we should force people to use it instead of plain ol' HTTP.


edwardloveall:
I dunno about this. You're introducing a lot of code here. Can you break it up into multiple commits?

gabebw:
Good call. I broke it up into 1 line/commit -- is it easier to read now?

edwardloveall:
Yes, much better. Thank you. :shipit:

adarsh:
aw_snap

PR #32, opened 2016-02-02
Recommend secure webring links

gabebw:
* When an HTTPS site links to the HTTP version of Hotline Webring, it won't send a Referer header
* Hotline Webring relies on the Referer header to know which site to add to the ring
* Linking to the HTTPS version of Hotline Webring ensures that we always get a Referer header, whether the linking site is using HTTPS or not


edwardloveall:
:shipit:

PR #33, opened 2016-02-02
Use created_at in atom feed entries

edwardloveall:
When a new Redirection is added to the webring, the first Redirection in the ring points to it. This changes that first Redirection's `updated_at` value.

In the atom feed, we were using `updated_at` to say when that entry was updated. While that's semantic, it would cause the first Redirection to continue to pop up in atom feeds because it looked like there was new information. No longer!

We now use `created_at` in that field, because for the purposes of the feed, we only want to see each Redirection when it's added, not "updated".

### DIFF
--- a/spec/requests/feed_requests_spec.rb
+++ b/spec/requests/feed_requests_spec.rb
@@ -36,6 +36,19 @@ RSpec.describe "Feed requests" do
       expect(first_entry["id"]).to eq(id)
       expect(first_entry["updated"]).to eq(updated)
     end
+
+    it "uses created_at instead of updated_at for <updated> value" do
+      created_at = 1.day.from_now
+      create(
+        :redirection,
+        created_at: created_at,
+        updated_at: created_at + 1.day
+      )
+
+      get feed_path
+
+      expect(first_entry["updated"]).to eq(created_at.iso8601)
+    end
   end
 
   def xml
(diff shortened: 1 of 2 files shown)


spec/requests/feed_requests_spec.rb
```diff
@@ -36,6 +36,17 @@
       expect(first_entry["id"]).to eq(id)
       expect(first_entry["updated"]).to eq(updated)
     end
+
+    it "uses created_at instead of updated_at for updated value" do
+      created_at = 1.day.from_now
+      create(:redirection,
```
gabebw: Formatting:

```ruby
create(
  :redirection,
  created_at: created_at,
  updated_at: 2.days.from_now,
)
```

Also, how about defining `updated_at` in terms of `created_at` to show that they're different? Maybe `updated_at: created_at + 3.days`?

spec/requests/feed_requests_spec.rb
```diff
@@ -36,6 +36,17 @@
       expect(first_entry["id"]).to eq(id)
       expect(first_entry["updated"]).to eq(updated)
     end
+
+    it "uses created_at instead of updated_at for updated value" do
```
gabebw: Maybe `<updated> value`?

edwardloveall: sure.

edwardloveall: Sorry that sounded mean. I just was lazy with capitalization but not with punctuation.

edwardloveall: Also, I'm super pissed at you.

spec/requests/feed_requests_spec.rb
edwardloveall: sure.

gabebw:
Beautiful :+1:

PR #36, opened 2016-02-06
Use Ruby 2.3.0

gabebw:
The performance gains are negligible, but sticking with the latest Ruby will
make later upgrades easier.


adarsh:
What about `.ruby-version` and/or `.travis.yml`?

gabebw:
Oh right. Fixed `.ruby-version` in 18bc6b2bb64c0afdcedfa08b9814aa848a8b2a9d

adarsh:
:sheep:

PR #38, opened 2016-02-07
Only repeat the site's name once in the header

gabebw:
Remove the underlines on the arrow links.


edwardloveall:
Looks good to me, pending circle

PR #39, opened 2016-02-21
Create a marquee-style banner to advertise t-shirts

edwardloveall:
Looks like this but moves right to left:
screen shot 2016-02-21 at 11 07 01 am

### DIFF
--- a/app/assets/stylesheets/_shirts.scss
+++ b/app/assets/stylesheets/_shirts.scss
@@ -1,0 +1,30 @@
+div.shirt-banner {
+  margin-bottom: $base-spacing;
+  overflow: hidden;
+  white-space: nowrap;
+
+  a {
+    @include animation(marquee 20.0s linear infinite);
+    color: white;
+    display: inline-block;
+    font-style: italic;
+    font-weight: bold;
+    padding-left: 100%;
+    text-decoration: none;
+    text-transform: uppercase;
+  }
+
+  @include keyframes(marquee) {
+    0% { @include transform(translate(0, 0)); }
+    100% { @include transform(translate(-100%, 0)); }
+  }
+
+  a:hover {
+    text-decoration: underline;
+  }
+
+  @include keyframes(marquee) {
+    0% { @include transform(translate(0, 0)); }
+    100% { @include transform(translate(-100%, 0)); }
+  }
+}
--- a/app/assets/stylesheets/_shirts.scss
+++ b/app/assets/stylesheets/_shirts.scss
@@ -14,11 +14,6 @@ div.shirt-banner {
     text-transform: uppercase;
   }
 
-  @include keyframes(marquee) {
-    0% { @include transform(translate(0, 0)); }
-    100% { @include transform(translate(-100%, 0)); }
-  }
-
   a:hover {
     text-decoration: underline;
   }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,3 +8,4 @@
 @import "refills/flashes";
 @import "landing-page";
 @import "mobile";
+@import "shirts";
--- a/app/views/homes/show.html.erb
+++ b/app/views/homes/show.html.erb
@@ -7,6 +7,10 @@
 </header>
 
 <section class="text">
+  <div class="shirt-banner">
+    <a href="http://teespring.com/hotlinewebring" target="_blank">T-shirts for a limited time!</a>
+  </div>
+
   <p>
     Do you long for a simpler time, when America was Online and the only person you
     could Ask was Jeeves? Hotline Webring is bringing that time back, with


app/assets/stylesheets/_shirts.scss
```diff
@@ -0,0 +1,21 @@
...
+    padding-left: 100%;
+    text-decoration: none;
+    text-transform: uppercase;
+  }
+}
+
+@include keyframes(marquee) {
+  0% { @include transform(translate(0, 0)); }
+  100% { @include transform(translate(-100%, 0)); }
+}
```
adarsh: Uh, why so fancy?

```
<marquee>This text will scroll from right to left</marquee>
```
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee

edwardloveall: From those docs:
>**Obsolete**
>This feature is obsolete. Although it may still work in some browsers, its use is discouraged since it could be removed at any time. Try to avoid using it.

edwardloveall: I guess I just didn't feel comfortable using the old element, even if it might work. If it works in all browsers though I'm not totally opposed.

adarsh: That's a bummer, man. It would have been _soooooo on brand_.

On Sunday, February 21, 2016, Edward Loveall <notifications@github.com>
wrote:

> In app/assets/stylesheets/_shirts.scss
> <https://github.com/gabebw/hotline-webring/pull/39#discussion_r53569461>:
>
> > +  a {
> > +    @include animation(marquee 20.0s linear infinite);
> > +    color: white;
> > +    display: inline-block;
> > +    font-style: italic;
> > +    font-weight: bold;
> > +    padding-left: 100%;
> > +    text-decoration: none;
> > +    text-transform: uppercase;
> > +  }
> > +}
> > +
> > +@include keyframes(marquee) {
> > +  0% { @include transform(translate(0, 0)); }
> > +  100% { @include transform(translate(-100%, 0)); }
> > +}
>
> I guess I just didn't feel comfortable using the old element, even if it
> might work. If it works in all browsers though I'm not totally opposed.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gabebw/hotline-webring/pull/39/files#r53569461>.
>


-- 


-- 
313.454.1515 (mobile)

gabebw:
Oh _hell yes_.

:+1: :100: :heart_eyes_cat:

app/assets/stylesheets/_shirts.scss
gabebw: We _could_ do a `marquee` element instead of `div.shirt-banner`, and style the `marquee`.

adarsh: Very on-brand. Think about it: what If someone views the source?

On Sunday, February 21, 2016, Gabe Berke-Williams <notifications@github.com>
wrote:

> In app/assets/stylesheets/_shirts.scss
> <https://github.com/gabebw/hotline-webring/pull/39#discussion_r53571620>:
>
> > +  a {
> > +    @include animation(marquee 20.0s linear infinite);
> > +    color: white;
> > +    display: inline-block;
> > +    font-style: italic;
> > +    font-weight: bold;
> > +    padding-left: 100%;
> > +    text-decoration: none;
> > +    text-transform: uppercase;
> > +  }
> > +}
> > +
> > +@include keyframes(marquee) {
> > +  0% { @include transform(translate(0, 0)); }
> > +  100% { @include transform(translate(-100%, 0)); }
> > +}
>
> We *could* do a marquee element instead of div.shirt-banner, and style
> the marquee.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gabebw/hotline-webring/pull/39/files#r53571620>.
>


-- 


-- 
313.454.1515 (mobile)

app/assets/stylesheets/_shirts.scss
```diff
@@ -0,0 +1,21 @@
...
+  white-space: nowrap;
+
+  a {
+    @include animation(marquee 20.0s linear infinite);
+    color: white;
+    display: inline-block;
+    font-style: italic;
+    font-weight: bold;
+    padding-left: 100%;
+    text-decoration: none;
```
gabebw: Is it obvious that it's a link?

edwardloveall: Eh... it doesn't have an underline (which I think makes it look slightly better), but it does scroll across the screen and it's bold and italic. It's not exactly subtle, but no, there is nothing that explicitly says "I'm a link" until you hover over it or click it. I'm feel about a 2/5 on this so if you feel strongly I'll add it!

gabebw: The arrows don't look like links either, so this is fine.

Maybe we should add underlines to everything later (in a separate PR)?

app/assets/stylesheets/_shirts.scss
edwardloveall: On one hand, we're emulating a very old (relatively) school concept, similar to how the `<marquee>` is old.

On the other hand we're using all sorts of modern (relatively) tech to do it (http 1.1, rails, heroku) unlike `<marquee>`.

I can see it both ways, but it seems like you are for the `<marquee>` option, so I'll change it. :)

gabebw:
:+1:

gabebw:
Oh, wait: I meant we should use `<marquee>` for Old School Reasons, but style it with the `transform` so it doesn't break in the future.

edwardloveall:
I don't understand how that would work. Without any style, the marquee moves by itself. Adding the `transform` style would make it move double.

gabebw:
> would make it move double

Oooh, sounds exciting!

JK, just follow your heart. I'm inclined towards your original non-marquee version, because as much as I love 90s nostalgia, I love following standards even more.

app/assets/stylesheets/_shirts.scss
edwardloveall: I am mostly pro-no underline because it looks cleaner. As long as there some other way to give affordance to the links. Maybe we could make all links white and change headers to a different color?

gabebw: Sure, whatever looks good. I trust you.

gabebw:
:+1: :+1: :tada: :sparkles: :100:

edwardloveall:
Oh right, I should merge this

PR #40, opened 2016-02-26
Show the slug in the Atom feed

gabebw:
This will make it easier to parse for anyone who writes, say, a best-selling iOS app.

### DIFF
--- a/app/views/feeds/show.atom.builder
+++ b/app/views/feeds/show.atom.builder
@@ -9,6 +9,7 @@ atom_feed do |feed|
     feed.tag!(:entry) do |entry|
       entry.id redirection.tag_uri
       entry.title "#{redirection.slug} joined #{title}"
+      entry.slug redirection.slug
       entry.updated redirection.created_at.iso8601
       entry.content redirection.url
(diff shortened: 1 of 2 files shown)


edwardloveall:
Would an iOS app rather have an API? What made you want to include it in the Atom feed?

There's no slug in [the Atom spec](http://www.ietf.org/rfc/rfc4287.txt). The closest fit would probably be the summary field, but even that is kinda weird.

gabebw:
You're right, I don't know what I was thinking. It's not even in the spec!

An API is a better idea if, say, @jnutting wants to build an app...

Closing without merging.

adarsh:
Webring app!?! Brave new world

On Friday, February 26, 2016, Gabe Berke-Williams <notifications@github.com>
wrote:

> Closed #40 <https://github.com/gabebw/hotline-webring/pull/40>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gabebw/hotline-webring/pull/40#event-567851414>.
>


-- 


-- 
313.454.1515 (mobile)

PR #45, opened 2016-05-06
Store the path for referrers with home directories

### LINKED ISSUE #44: Redirect to the path (not just the host) for accounts like http://tilde.town/~vilmibm/
Redirecting to tilde.town in these cases is very different from redirecting to http://tilde.town/~vilmibm/, so we should redirect to http://tilde.town/~vilmibm/.

gabebw:
Until now, the webring always normalized referrers to the hostname, with no path (like `tilde.club`). Here's a case where that breaks down.

A classic method of user management for shared servers like http://tilde.club/ is to store each user's content under a top-level path like `/~gabe`.

Gabe would be at tilde.club/~gabe, Edward would be at tilde.club/~edward, and so on. One person curates `/~gabe`, another person curates `/~edward`, and a third person curates the tilde.club homepage itself.

In the case of shared servers, when someone adds webring links to a page on (e.g.) `tilde.club/~gabe/webring.html`, the webring stores `tilde.club`. We should instead respect each user's individuality and store that site's URL as `tilde.club/~gabe`.

Fixes #44.

### DIFF
--- a/app/services/redirection_creation.rb
+++ b/app/services/redirection_creation.rb
@@ -12,7 +12,7 @@ class RedirectionCreation
     redirection = Redirection.new(slug: slug)
 
     if referrer.present?
-      redirection.url = referrer_hostname
+      redirection.url = normalized_referrer
       Ring.new(redirection).link
       redirection
     end
@@ -22,8 +22,19 @@ class RedirectionCreation
 
   attr_reader :referrer, :slug
 
+  def normalized_referrer
+    if referrer_uri.path.start_with?("/~")
+      referrer_hostname + File.dirname(referrer_uri.path)
+    else
+      referrer_hostname
+    end
+  end
+
   def referrer_hostname
-    referrer_uri = URI.parse(referrer)
     "#{referrer_uri.scheme}://#{referrer_uri.hostname}"
+  end
+
+  def referrer_uri
+    @_referrer_uri ||= URI.parse(referrer)
   end
 end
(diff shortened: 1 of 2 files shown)


edwardloveall:
Been thinking about this, sorry I haven't responded. What if we just use the full URL instead of hostname or hostname/~username? The redirection is created when the user clicks on the link. So for example, in your test, if the referrer is `http://tilde.club/~ford/something.html`, then that means the link is on that page. If redirect to `http://tilde.club/~ford/`, we're in the same boat as we are now.

This also makes it possible to have anyone use any URL. Is there a reason we used only hostnames in the past? Normalization?

gabebw:
>  So for example, in your test, if the referrer is http://tilde.club/~ford/something.html, then that means the link is on that page. If redirect to http://tilde.club/~ford/, we're in the same boat as we are now.

Well, sort of. I think that, based on the naming convention, `/~ford` is likely to be a hub of ford's activity and `something.html` should be accessible from there.

> What if we just use the full URL?  Is there a reason we used only hostnames in the past? Normalization?

Yes, I think we were worried about double entries for a site, both malicious and unintentional. For example, I have webring links on every blog post on gabebw.com, but I don't want 1 webring entry per blog post. Normalizing to the host prevents that.

gabebw:
I should say that I'm definitely not against just using the raw referer. But I can't think of anything offhand that solves my blog-post-webring-link problem.

edwardloveall:
Ah yes, thank you for reminding me. Another solution could be redirects to any arbitrary domain but limited one per hostname. Also not a silver bullet.

What if we just put this one in manually? It's not as if we're getting 100 requests a day or something. We could just deal with it on a case by case basis.

gabebw:
> What if we just put this one in manually?

I did actually change it manually: https://hotlinewebring.club/

> We could just deal with it on a case by case basis.

I think that's the best solution here.

PR #47, opened 2016-05-12
Add a test for visiting a High Voltage page

gabebw:
On production, starting the server immediately fails with this error:

    undefined method `hide_action' for HighVoltage::PagesController:Class

This is because `hide_action` was removed from controllers in Rails 5.

This error does _not_ occur in development or test because those environments have `config.eager_load` set to `false`, so the `HighVoltage::PagesController` code is never loaded, and thus doesn't fail.

The test added in this commit exercises the `HighVoltage::PagesController` code, and fails without upgrading high_voltage.

To fix this issue, upgrade to high_voltage 3.0.0, which supports Rails 5: https://github.com/thoughtbot/high_voltage/blob/master/NEWS.md

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -94,7 +94,7 @@ GEM
     globalid (0.3.6)
       activesupport (>= 4.1.0)
     hashdiff (0.2.2)
-    high_voltage (2.4.0)
+    high_voltage (3.0.0)
     i18n (0.7.0)
     jquery-rails (4.0.5)
       rails-dom-testing (~> 1.0)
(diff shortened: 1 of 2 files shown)


spec/features/user_visits_static_page_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.feature "User visits static page" do
```
gabebw: I don't love this spec description.

edwardloveall: What about `User visits localhost page`?

edwardloveall:
Suggestion for your test name, otherwise looks good 👍

PR #48, opened 2016-05-21
Add a method to remove a site from the webring

gabebw:
I know, I know: who would want such a thing?!

The `unlink` method runs in a transaction so that removing a Redirection and re-linking its neighbors is an atomic operation. In other words, we can't remove a Redirection, run into an error, and then leave a hole in the ring.

### DIFF
--- a/app/models/redirection.rb
+++ b/app/models/redirection.rb
@@ -26,4 +26,14 @@ class Redirection < ActiveRecord::Base
 
     "tag:#{UNCHANGING_HOSTNAME},#{date}:#{url}"
   end
+
+  def unlink
+    previous_redirection = previous
+    next_redirection = self.next
+
+    self.class.transaction do
+      destroy!
+      previous_redirection.update!(next: next_redirection)
+    end
+  end
 end
--- a/spec/models/redirection_spec.rb
+++ b/spec/models/redirection_spec.rb
@@ -83,4 +83,27 @@ RSpec.describe Redirection do
       expect(redirection.tag_uri).to eq(id)
     end
   end
+
+  describe "#unlink" do
+    it "destroys the redirection" do
+      redirection = create(:redirection)
+
+      redirection.unlink
+
+      expect(Redirection.exists?(id: redirection.id)).to be_falsey
+    end
+
+    it "links the redirection's previous to its next" do
+      gabe = Redirection.find_by!(slug: "gabe")
+      edward = Redirection.find_by!(slug: "edward")
+      redirection = create(:redirection)
+
+      redirection.unlink
+
+      edward.reload
+      gabe.reload
+      expect(edward.next).to eq gabe
+      expect(gabe.next).to eq edward
+    end
+  end
 end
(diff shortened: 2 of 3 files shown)


spec/models/redirection_spec.rb
```diff
@@ -83,4 +83,32 @@
...
     end
   end
+
+  describe "#unlink" do
+    it "destroys the redirection" do
+      redirection = create(:redirection)
+
+      redirection.unlink
+
+      expect(Redirection.find_by(id: redirection.id)).to be_nil
```
gabebw: This isn't as pretty as I'd like, but:

* `expect(Redirection.count).to eq 0` fails because we always have 2 seeded redirections, for Edward and Gabe, and
* `expect(Redirection.count).to eq 2` is just letting a [Mystery Guest](https://robots.thoughtbot.com/mystery-guest) wipe their muddy shoes all over our nice clean house

app/models/redirection.rb
```diff
@@ -26,4 +26,14 @@ def tag_uri
 
     "tag:#{UNCHANGING_HOSTNAME},#{date}:#{url}"
   end
+
+  def unlink
+    previous_redirection = previous
+    next_redirection = self.next
```
edwardloveall: Would plain `next` not work? Why the `self`?

spec/models/redirection_spec.rb
edwardloveall: Could also do:

```ruby
redirection_count = Redirection.count

# do stuff

expect(Redirection.count).to eq(redirection_count - 1)
```

I kind of like your nil check though. What about something like:

```ruby
redirection = create(:redirection)

redirection.unlink
redirection.reload

expect(redirection).to be_nil
```

Yet another alternative could be a `deleted?` predicate on `Redirection` and then we could use `be_deleted`

edwardloveall:
Good stuff. Test might be able to be cleaned up but otherwise looks good.

app/models/redirection.rb
gabebw: `next` is a keyword in Ruby, so it must be `self.next`. I've tried to think of something better that doesn't require the `self.`, but I like the `next`/`previous` names enough that I don't mind the `self.`.

spec/models/redirection_spec.rb
gabebw: Unfortunately, `redirection.reload` with a deleted Redirection raises an `ActiveRecord::RecordNotFound` error.

We could add `deleted?`, but I'm wary of adding application code that's only used in a test. What would go in the `deleted?` method?

gabebw: How about:

```ruby
expect(Redirection.exists?(id: redirection.id)).to eq false
```

app/models/redirection.rb
edwardloveall: Oh right.

spec/models/redirection_spec.rb
edwardloveall: >`redirection.reload` with a deleted Redirection raises an `ActiveRecord::RecordNotFound` error.

Good point! What about:

```ruby
expect { redirection.reload }.to raise_exception(ActiveRecord::RecordNotFound)
```

edwardloveall: If you don't like that, your solution is fine. This is Not That Big Of A Deal™

P.S. `be_falsey` is a good matcher too.

PR #50, opened 2016-07-08
Use Rails 5.0.0

gabebw:
We were using Rails 5.0.0.rc1.

* Instead of the `quiet_assets` gem, set `config.assets.quiet = true`: https://github.com/rails/sprockets-rails/pull/355
* Remove `coffee-rails`, which wasn't being used


edwardloveall:
Love it. Ship it.

PR #55, opened 2016-08-22
Self-host the bitmap of a Clumber Spaniel

gabebw:
The external link was broken. To prevent this from happening again, self-host the spaniel.


adarsh:
👍  That's a big image. Maybe run an optimizer on it?

gabebw:
BMPs are pretty huge, by default. I don't think they're optimizable.

gabebw:
I think having an image load slowly is on-brand and conveys our 90s #aesthetic, too.

adarsh:
Oh totally. Hopefully from the bottom to the top, rasterizing.



-- 
313.454.1515 (mobile)

gabebw:
That bottom-to-top rendering is why I love bitmaps. It's like magic!

PR #59, opened 2016-09-28
Remove JavaScript

gabebw:
* `application.js` only pulls in jQuery, which isn't used anywhere
* The `_analytics` partial loads Segment JS, which we don't use


edwardloveall:
This may need to come back if we ever add a form on the website, but good call for now.

PR #60, opened 2016-09-28
Add link_audit rake task

edwardloveall:
This taks should help us to maintenance on the webring's members. It
checks each redirection for proper next and previous links. It downloads
the page and scans the raw html for the proper urls. It doesn't check
for actual links because some sites (looking at you @jnutting) don't use
normal `<a href="...">` links.

This also found other issues while I was testing it like sites that have
migrated to https because the html only returned the redirection (and
therefore didn't include the links).

### DIFF
--- a/lib/tasks/link-audit.rake
+++ b/lib/tasks/link-audit.rake
@@ -1,0 +1,27 @@
+desc "Determine if all websites have links"
+task link_audit: :environment do
+  def green(text)
+    puts "\e[32m#{text}\e[0m"
+  end
+
+  def red(text)
+    puts "\e[31m#{text}\e[0m"
+  end
+
+  Redirection.find_each do |redirection|
+    url = URI(redirection.url)
+    html = Net::HTTP.get(url)
+    next_link = "hotlinewebring.club/#{redirection.slug}/next"
+    prev_link = "hotlinewebring.club/#{redirection.slug}/previous"
+
+    missing_links = []
+    missing_links << "next" if !html.include?(next_link)
+    missing_links << "prev" if !html.include?(prev_link)
+
+    if missing_links.empty?
+      green("#{redirection.url} is all good")
+    else
+      red("#{redirection.url} is missing #{missing_links.join(' and ')}")
+    end
+  end
+end


lib/tasks/link-audit.rake
```diff
@@ -0,0 +1,22 @@
...
+    html = Net::HTTP.get(url)
+    next_link = "hotlinewebring.club/#{redirection.slug}/next"
+    prev_link = "hotlinewebring.club/#{redirection.slug}/previous"
+
+    missing_links = []
+    missing_links << "next" if !html.include?(next_link)
+    missing_links << "prev" if !html.include?(prev_link)
+
+    if missing_links.empty?
+      print "\e[32m"
```
gabebw: Could you make a `green` function (or whatever color this is), like so?

```ruby
def green(text)
  print "\e[32m"
  puts "#{redirection.url} is all good"
  print "\e[0m"
end
```

I personally enjoy putting the "reset color" code in the function rather than printing it out after the `if`.

lib/tasks/link-audit.rake
```diff
@@ -0,0 +1,22 @@
...
+      print "#{redirection.url} is all good"
+    else
+      print "\e[31m"
```
gabebw: Same here for (I'm guessing) `red`.

lib/tasks/link-audit.rake
```diff
@@ -0,0 +1,22 @@
+desc "Determine if all websites have links"
+task link_audit: :environment do
+  Redirection.find_each do |redirection|
+    url = URI(redirection.url)
```
gabebw: What about calling this `uri` instead of `url`?

edwardloveall: Yeah I kind of use those interchangeably. I guess URI is more generic in that it's _some_ form of identifier, and URLs are specifically for the web. So I am technically dealing with URLs even if the Net::HTTP requires a URI. Anyway, yeah I'll change it.

gabebw:
Thank you for building this, this should make maintenance a lot easier!

PR #65, opened 2018-01-02
Use the latest version of Ruby

gabebw:


gabebw:
bundler-audit is failing the build because we have an old version of Nokogiri.

Unfortunately, upgrading Nokogiri to a safe version requires Rails 5.1 (which is what I'm doing in #66). I'll combine #66 into this one and close #66.

gabebw:
DatabaseCleaner might be needed for CircleCI.  I'll try building _without_ removing DatabaseCleaner.

gabebw:
I decided not to remove Database Cleaner in this PR. Instead, I opened #67 to do so in the future.

gabebw:
@edwardloveall Review me plz 🙏

PR #66, opened 2018-01-02
Update Rails to the latest version

gabebw:
Update `simple_form` and `dotenv-rails` too, since they rely on version of Rails <5.1.

Also, remove DatabaseCleaner, which [is not needed in Rails 5.1](https://github.com/thoughtbot/suspenders/pull/864).

**NOTE**: #65 should be merged before this PR.


gabebw:
Closing in favor of #65.

PR #71, opened 2018-12-12
Remove capybara-webkit

gabebw:
It no longer installs on macOS Sierra (the [wiki] recommends checking out an old version of Homebrew to install Qt5.5), making it difficult to run tests on a new computer.

In addition, we don't have any Javascript tests.

Thus, the easiest thing is just to not use capybara-webkit.

[wiki]: https://github.com/thoughtbot/capybara-webkit/wiki/Installing-Qt-and-compiling-capybara-webkit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,6 @@ group :development, :test do
 end
 
 group :test do
-  gem "capybara-webkit"
   gem "database_cleaner"
   gem "formulaic"
   gem "launchy"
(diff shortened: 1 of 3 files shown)


gabebw:
@edwardloveall Please review

gabebw:
I suspect the tests will fail due to vulnerable gems, but will otherwise pass. (That's what happens on my machine.

gabebw:
I opened #72 to update the vulnerable gems.

PR #73, opened 2018-12-12
Remove DatabaseCleaner

gabebw:
It's not needed in Rails 5.1: https://github.com/thoughtbot/suspenders/pull/864


edwardloveall:
Hmmm looking at the spec failure, maybe we need to set something so tests get wrapped in transactions?

gabebw:
Yeah, everything is failing 😞 

Even with `config.use_transactional_fixtures` set to `true`, everything is failing.

spec/rails_helper.rb
```diff
@@ -13,10 +13,13 @@ module Features
 end
 
 RSpec.configure do |config|
+  config.before(:suite) do
+    Rails.application.load_seed
```
gabebw: @edwardloveall FYI I also needed to copy this line from the now-deleted `database_cleaner.rb`.

PR #75, opened 2018-12-12
Upgrade to Ruby 2.5.3

gabebw:
2.5.3 is the latest version of Ruby: https://www.ruby-lang.org/en/downloads/


.ruby-version
```diff
@@ -1 +1 @@
-2.4.2
+2.5.3
```
edwardloveall: What do you think abut using `asdf`'s convention i.e. `.tools-version`?

PR #77, opened 2018-12-12
Use CircleCI 2.0

gabebw:
As you can see [in
Circle](https://circleci.com/gh/gabebw/hotline-webring), the `circleci-20-test` branches now take ~20 seconds total, while branches still on CircleCI 1.0 (like `master`) took 1 minute and 30 seconds.

I kept all of my commits here so you can follow what I did. I'll squash when I merge.

Thank you to https://github.com/thoughtbot/suspenders/pull/934 and https://circleci.com/sunset1-0/ for a lot of help!


.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+    shell: /bin/bash --login
+    environment:
+      CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
+      CIRCLE_TEST_REPORTS: /tmp/circleci-test-results
+    docker:
+    - image: circleci/ruby:2.5.3-node
+      environment:
+        RAILS_ENV: test
+        PGHOST: localhost
+        PGUSER: hotlinewebring
```
gabebw: This can be anything; it just needs to match `POSTGRES_USER` below.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+    - image: circleci/postgres:10.6
+      environment:
+        POSTGRES_USER: hotlinewebring
+        POSTGRES_DB: hotlinewebring_test
+        POSTGRES_PASSWORD: ""
+
+    steps:
+      - checkout
```
gabebw: This is a magic command that checks out the code.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+          command: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
+      - run:
+          name: Create .env
+          command: cp .sample.env .env
+
+      # Restore the dependency cache
+      - restore_cache:
+          keys:
+          - hotlinewebring-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
+          - hotlinewebring-
```
gabebw: We could probably remove this line, which is a fallback place to look for the dependency cache.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  build:
+    working_directory: ~/gabebw/hotline-webring
+    parallelism: 1
+    shell: /bin/bash --login
+    environment:
+      CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
+      CIRCLE_TEST_REPORTS: /tmp/circleci-test-results
+    docker:
+    - image: circleci/ruby:2.5.3-node
```
gabebw: We actually don't need Node, I think.

gabebw: We do need Node for `execjs`, which is required by `autoprefixer-rails`.

PR #79, opened 2018-12-12
Correctly show failing tests

gabebw:
You can see that the first commit, which removed `db:test:prepare` (which we do need), has a bad Circle CI output and doesn't summarize the failing tests at the top: https://circleci.com/gh/gabebw/hotline-webring/280

However, the latest commit, which stores test results correctly, does summarize the failing tests: https://circleci.com/gh/gabebw/hotline-webring/283

(I will remove the failing commit once this PR is approved.)

Using `path: $CIRCLE_TEST_REPORTS` made it look at a directory literally called `$CIRCLE_TEST_REPORTS`. It didn't use the value of the environment variable. Therefore, I now hardcode it as a string.

According to  https://circleci.com/docs/2.0/configuration-reference/#store_test_results, we should store it to `test-results/rspec/results.xml`, and save `test-results`, so now we do that instead of using environment variables.

I'm not sure we have any artifacts, so I removed that entirely.

@edwardloveall

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,9 +4,6 @@ jobs:
     working_directory: ~/gabebw/hotline-webring
     parallelism: 1
     shell: /bin/bash --login
-    environment:
-      CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
-      CIRCLE_TEST_REPORTS: /tmp/circleci-test-results
     docker:
     - image: circleci/ruby:2.5.3-node
       environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -49,19 +46,12 @@ jobs:
           command: dockerize -wait tcp://localhost:5432 -timeout 1m
 
       - run: bin/setup
-      - run: bundle exec bin/drake db:test:prepare
```
gabebw: This is from the first commit, just so we have failing tests to check if Circle displays those failing tests correctly.

PR #86, opened 2018-12-16
Update to Rails 5.2.2

gabebw:
The biggest change aside from Rails is probably moving `simple_form` from 3.5.0 to 4.0.0. However, we don't have any forms in our app, so it can't affect us.

@edwardloveall


edwardloveall:
What do you think about removing simple form?

gabebw:
I agree, let's remove `simple_form`. Please hold.

gabebw:
@edwardloveall I opened a new PR to remove simple_form in #100

gabebw:
simple_form has been removed, and I force-pushed this branch.

PR #89, opened 2018-12-16
FactoryGirl is now named FactoryBot

gabebw:
It was renamed in version 4.8.2: https://github.com/thoughtbot/factory_bot/blob/master/NEWS

This also fixes a small warning from the new FactoryBot version:

```
DEPRECATION WARNING: Static attributes will be removed in FactoryBot
5.0. Please use dynamic attributes instead by wrapping the attribute
value in a block:

  next_id { -1 }
```


gabebw:
@edwardloveall

PR #103, opened 2018-12-16
Use a CircleCI cache across different branches

gabebw:
Previously, builds would start a new cache in these cases:

* when building a new branch for the first time (since the branch name changed)
* when `Gemfile.lock` changes

However, the cache only saves Gemfile-related paths. Therefore, the cache is OK to use for new branches, as long as the `Gemfile.lock` is the same.

After this commit, a build can use a different branch's cache as long as they're using the same `Gemfile.lock`. This should significantly speed up the first build on new branches.

I've also changed the name of the cache to include `gemfile`, so that if we add other caches in the future, it doesn't get confusing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
       # Restore the dependency cache
       - restore_cache:
           keys:
-          - hotlinewebring-{{ .Branch }}-{{ checksum "Gemfile.lock" }}
+          - hotlinewebring-gemfile-{{ checksum "Gemfile.lock" }}
 
       - run:
           name: Bundle Install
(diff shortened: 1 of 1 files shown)


gabebw:
An additional benefit to the changes in this PR: it speeds up the master branch builds as well. Previously, master had an existing cache for its branch, but NOT its gemfile. So when a branch changes a gemfile, master builds were taking a long time again. But now, the master branch can use the same cache from the branch that changed the gemfile, and we only pay the cost of `bundle install` once, on the branch that changed the gemfile.

You can see this happening [on the master branch in CircleCi](https://circleci.com/gh/gabebw/hotline-webring/tree/master); branches took 1 to 3 minutes until #347, which took 30 seconds.

PR #111, opened 2018-12-18
Bring back Capybara gem so feature specs run

gabebw:
In PR #105, we removed Capybara because it's not in our Gemfile explicitly and #105 removed the last gem depending on Capybara.

Because of this, none of our feature specs run:

```
1) User visits homepage sees list of redirections
  # Feature specs require the Capybara
  #  (http://github.com/jnicklas/capybara) gem, version 2.2.0 or later.
  #  We recommend version 2.4.0 or later to avoid some deprecation
  #  warnings and have support for `config.expose_dsl_globally =
  #  false`.
  # ./spec/features/user_visits_homepage_spec.rb:4
```

Therefore, bring back Capybara to run feature specs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ group :development, :test do
 end
 
 group :test do
+  gem "capybara"
   gem "launchy"
   gem "rspec_junit_formatter"
   gem "shoulda-matchers"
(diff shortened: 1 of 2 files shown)


gabebw:
I wish we could make RSpec fail those specs. Oh well.

PR #113, opened 2018-12-19
Prefer `heroku local` to Foreman

gabebw:
`heroku local` has been the preferred solution for a while now.

The `bin/setup` script now requires you to install `heroku`.


gabebw:
Related Suspenders PR with some more context: https://github.com/thoughtbot/suspenders/pull/655

Issue #58, opened 2016-09-28
The problem with how hostnames are added


edwardloveall:
The way it currently works is when a link is incoming, we normalize the url to be just the hostname and store that as the URL to redirect to. This has a couple of problems:
- If the next/prev links are on a subpage, it's hard to find the links
- If multiple users from a single site want to join the webring, they can't without manual intervention on our part.

However, if we switch to using the full referrer, we have another set of problems:
- If someone has the webring links as part of their website's global template, they could trigger a new redirection to be added.

Example: http://nuthole.com and http://nuthole.com/blog/2016/05/03/jack-shirt/ 

This can be fixed by only allowing one link per hostname, but that removes the possibility of having multiple members from a single site (like tilde.town).

Another option: an application form like [cuties.online](http://cuties.online/?Join)

This is a little bit of work to build (and not as slick) but would fix both of these problems. A user would pick what url to redirect to and we wouldn't have any accidentals adds.

Putting this out there to consider options.

gabebw:
I'm leaning towards an application form (probably a Google form that submits to both of our email addresses).

Reasons:

* no edge cases
* impossible to game
* no more accidentally adding people and asking them if they meant to do that
* humans can handle the tilde.town cases
* I don't think filling out the form will decrease participation _that_ much; people are highly motivated to join our exclusive, cool webring

edwardloveall:
Sounds good. I'd vote for a non-google form. I think they look pretty crappy. If it's the only free option, I'll code one up.

gabebw:
These are the options:

* Google forms (free), doesn't look good
* Formkeep ($29/month), styleable
* [Typeform](https://www.typeform.com/), ([free](https://www.typeform.com/pricing/)), looks good

Looks like Typeform is the winner.

gabebw:
As of December 20, 2018, we store the whole original referring URL (#115) in addition to the hostname, which might (with some rare manual intervention on our part) make this issue go away.

>If the next/prev links are on a subpage, it's hard to find the links

We check each new link that's added to make sure this isn't an issue (which honestly I enjoy, since it means the webring is growing!), and now that we have the full URL we can fix it if necessary.

>If multiple users from a single site want to join the webring, they can't without manual intervention on our part.

I think the reasoning above about checking each new URL fixes this too.

>However, if we switch to using the full referrer, we have another set of problems:
>
> * If someone has the webring links as part of their website's global template, they could trigger a new redirection to be added.

[...]

PR #115, opened 2018-12-21
Store the whole original referring URL

gabebw:
We currently store only the hostname of the HTTP referrer (e.g. `gabebw.com/posts/2018/whatever` is saved as just `gabebw.com`).

This is a problem for pages on free sites, where the homepage for a given user is not the root hostname but a subdirectory (e.g. `freepages.rootsweb.com/~gabebw`).

This is not a fix for issue #58, but it does allow us to manually fix that issue if it recurs.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2015_12_17_165101) do
+ActiveRecord::Schema.define(version: 2018_12_20_204317) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2015_12_17_165101) do
+ActiveRecord::Schema.define(version: 2018_12_20_204317) do
```
gabebw: Almost exactly 3 years between database updates. Pretty good!

spec/services/redirection_creation_spec.rb
```diff
@@ -15,6 +15,17 @@
       expect(redirection.url).to eq referrer_hostname
     end
 
+    it "stores the original unchanged referrer" do
+      referrer_hostname = "https://cool.example.com"
+      full_referrer = "#{referrer_hostname}/something/else"
+      slug = "cool-slug"
```
edwardloveall: 🐌

PR #117, opened 2018-12-21
Configure rack-timeout with env variable

gabebw:
The `Rack::Timeout.timeout=` method no longer exists. It was removed in rack-timeout 0.5, which we upgraded to in commit 003248f9.

Relevant Suspenders PR: https://github.com/thoughtbot/suspenders/pull/920


gabebw:
I set `RACK_TIMEOUT_SERVICE_TIMEOUT` to `10` in both staging and production.

PR #177, opened 2019-06-10
Rescue errors and detect redirects in link audit

gabebw:
When trying to run `Net::HTTP.get_response` on a website that is not online (like http://alexfox.me, in our webring), it throws a `SocketError`.

Now we rescue that error and print an error message, rather than crashing the audit.

I also improved the redirection detection to now support `Net::HTTPMovedPermanently` (a 301 Redirect) as well.

### DIFF
--- a/lib/tasks/link-audit.rake
+++ b/lib/tasks/link-audit.rake
@@ -15,7 +15,7 @@ task link_audit: :environment do
       case response
       when Net::HTTPSuccess then
         response.body
-      when Net::HTTPRedirection then
+      when Net::HTTPRedirection, Net::HTTPMovedPermanently then
         fetch(response["location"])
       end
     rescue SocketError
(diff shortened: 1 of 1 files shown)


gabebw:
I thought about refactoring the code in the `link_audit` Rake task to its own class, but ultimately decided not to do so in this PR.

Advantages of refactoring:

* Easy to unit-test
* Rake task becomes something simple like `LinkAudit.run`

Disadvantages in this PR:

* It'd make reviewing this PR pretty dang difficult!

So, that's something that we might want to do in the future, but didn't seem like a good idea in this PR.

lib/tasks/link-audit.rake
```diff
@@ -10,28 +10,37 @@ task link_audit: :environment do
 
   def fetch(url_string)
     url = URI(url_string)
-    response = Net::HTTP.get_response(url)
```
gabebw: This PR is much easier to read if you ignore whitespace: https://github.com/gabebw/hotline-webring/pull/177/files?w=1

PR #179, opened 2019-06-11
Allow encoded slashes

gabebw:
As you can see by doing `curl https://alexeyzabelin.com`, this HTML works just fine in the browser:

```
<a href="https:&#x2F;&#x2F;hotlinewebring.club&#x2F;alexey&#x2F;next">next</a>
```

`&#x2F` is the hexadecimal encoding of `/`.

Therefore, it should "count" as a real HLWR link.

I have updated the link checker to allow both versions of the forward slash, regular and encoded.

### DIFF
--- a/lib/tasks/link-audit.rake
+++ b/lib/tasks/link-audit.rake
@@ -27,12 +27,17 @@ task link_audit: :environment do
     html = fetch(redirection.url)
 
     if html
-      next_link = "hotlinewebring.club/#{redirection.slug}/next"
-      prev_link = "hotlinewebring.club/#{redirection.slug}/previous"
+      # Surprisingly, this works just fine:
+      # <a href="https:&#x2F;&#x2F;hotlinewebring.club&#x2F;alexey&#x2F;next">next</a>
+      # Therefore, we should allow encoded slashes.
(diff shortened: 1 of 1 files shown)


lib/tasks/link-audit.rake
```diff
@@ -27,12 +27,17 @@ task link_audit: :environment do
     html = fetch(redirection.url)
 
     if html
-      next_link = "hotlinewebring.club/#{redirection.slug}/next"
-      prev_link = "hotlinewebring.club/#{redirection.slug}/previous"
+      # Surprisingly, this works just fine:
+      # <a href="https:&#x2F;&#x2F;hotlinewebring.club&#x2F;alexey&#x2F;next">next</a>
+      # Therefore, we should allow encoded slashes.
+      slash = %r{(/|&#x2F;)}
```
gabebw: Hmm, would this be clearer than what's there?

```ruby
slash = Regexp.union(['/', '&#x2f;'])
```

(I'm pretty sure the parentheses are unnecessary.)

edwardloveall: Yup! I like this better.

gabebw: Great, just made that change to use `Regexp.union` instead.

PR #181, opened 2019-06-13
Factor the link audit code into its own class

gabebw:
By doing this, I discovered a bug: http://alexeyzabelin.com has `&#x2F;` (note the uppercase "F") but our old code was checking for a lowercase "f". Now we check case-insensitively.

### DIFF
--- a/app/models/link_audit.rb
+++ b/app/models/link_audit.rb
@@ -1,0 +1,50 @@
+class LinkAudit
+  # Surprisingly, this works just fine:
+  # <a href="https:&#x2F;&#x2F;hotlinewebring.club&#x2F;alexey&#x2F;next">next</a>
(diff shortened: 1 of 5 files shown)


app/models/link_audit.rb
```diff
@@ -0,0 +1,50 @@
+class LinkAudit
```
edwardloveall: What do you think about calling this `LinkFinder` since that's all it's concerned with?

gabebw: I like it! Renamed to `MissingLinkFinder` in 48e0b737b0cf83ed0e61bfab00806e4d896fd9df.

edwardloveall: Even better!

PR #193, opened 2019-07-11
Update to ruby 2.6.3 (and some gems along with it)

edwardloveall:
About hashdiff:
This should remove this deprecation warning that is printed when booting
a Rails/Rspec process:
> The HashDiff constant used by this gem conflicts with another gem of a
> similar name.  As of version 1.0 the HashDiff constant will be
> completely removed and replaced by Hashdiff.  For more information see
> liufengyun/hashdiff#45.


Gemfile
```diff
@@ -32,6 +32,8 @@ end
 
 group :test do
   gem "capybara", "~> 3.0"
+  # TEMP: list hashdiff explicitly in Gemfile to force 1.0.0.beta1; remove once 1.0.0 is released
```
gabebw: Thank you for this comment 🙇

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.6.3
```
gabebw: Nice, I use asdf too.

Gemfile
edwardloveall: I am not watching the hashdiff project releases so hopefully I'll remember why (I have this same change on my portfolio) and update the Gemfiles

PR #258, opened 2020-02-24
Upgrade to Rails 6 (but I have some questions)

gabebw:
I changed to Rails 6 in the Gemfile and ran `rails app:update` and kept what seemed good:

* Support for new `rails dev:cache` command
* Support for `rails restart` with Puma
* Disable Action Mailer's fragment caching (which I think is new in Rails 6?), see section 2.4.1 here: https://guides.rubyonrails.org/action_mailer_basics.html#mailer-views
* Helpful config settings like `config.assets.quiet`, and support for tagged logging in production (`config.log_tags`, `config.logger`)


app/models/redirection.rb
```diff
@@ -1,5 +1,5 @@
 class Redirection < ActiveRecord::Base
-  belongs_to :next, class_name: "Redirection"
+  belongs_to :next, class_name: "Redirection", optional: true
```
gabebw: Without this line, this broke: https://github.com/gabebw/hotline-webring/blob/12bf532790761a282fbabf30897244af2182b0c1/db/seeds.rb#L2-L6

or this broke:

https://github.com/gabebw/hotline-webring/blob/12bf532790761a282fbabf30897244af2182b0c1/spec/factories.rb#L5

I don't know what changed in Rails 6 that makes this break, since I thought Rails 5 defaulted to `optional: false`, which I'd think would make this break.

Overview:

* Setting `next_id` to a fake value like `-1` breaks the validation here
* I tried `validates :next, presence: true, on: :update` so that creation would at least work, but that broke this:

https://github.com/gabebw/hotline-webring/blob/12bf532790761a282fbabf30897244af2182b0c1/app/models/ring.rb#L12

Basically, we need a way to tell Rails that yes, it's bad data...but we're going to fix it immediately.

Since I don't think we can do that, I just added this setting and will rely on tests.

I'm very open to hearing about a better way to to this!

config/environments/production.rb
```diff
@@ -3,18 +3,43 @@
   config.eager_load = true
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
+
+  # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
```
gabebw: We can just delete the comments here - I thought we might want them someday (e.g. `config.assets.css_compressor`), but happy to delete them (or uncomment them) too.

config/environments/test.rb
```diff
@@ -5,8 +5,10 @@
   config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
+  config.cache_store = :null_store
```
gabebw: Why not turn this off, right?

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
```
gabebw: I don't know what this does. Maybe just delete it?

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,28 @@
+# Be sure to restart your server when you modify this file.
```
gabebw: I doubt we need a CSP. Should I just delete it?

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
+# Be sure to restart your server when you modify this file.
```
gabebw: We should make an issue to flip these defaults, once this PR is merged.

spec/requests/feed_requests_spec.rb
```diff
@@ -7,7 +7,7 @@
 
       get feed_path
 
-      expect(response.content_type).to eq(Mime::Type.lookup_by_extension(:atom))
+      expect(response.media_type).to eq(Mime::Type.lookup_by_extension(:atom))
```
gabebw: To fix this warning:

> DEPRECATION WARNING: Rails 6.1 will return Content-Type header without modification. If you want just the MIME type, please use `#media_type` instead.

config/environments/production.rb
gabebw: @edwardloveall Does your 👍 mean "delete the comments in `config/environments/*.rb`"?

edwardloveall: It does! Sorry still haven't had a chance to look at the whole thing.

app/models/redirection.rb
edwardloveall: Hmmm yeah that's strange. My research turned up the same thing you found. I don't _love_ it being optional, but I think our test coverage is good enough that it's okay to leave it optional.

config/environments/test.rb
edwardloveall: [It defaults to `:file_store`](https://guides.rubyonrails.org/configuring.html). I don't know exactly what that will do, but I think we don't want any caching in tests so I would lean to say let's keep it `:null_store`

config/initializers/application_controller_renderer.rb
edwardloveall: Huh, yeah. If you delete it and everything seems fine then blow it away.

config/initializers/content_security_policy.rb
edwardloveall: Delete away.

PR #263, opened 2020-03-04
Update the valid hosts and TLDs based on the IANA list

WilHall:
👋 I was setting up a new `.dev` domain and noticed an issue getting it linked with hotline webring. This PR addresses that issue, and improves the accuracy of the host validator according to IANA's list of reserved domains.

This change addresses a few issues with the host validator:
 - `.dev` is now (as of 2019) a valid TLD (https://domains.google/tld/dev/)
 - IANA has a list of special use domain names which includes additional
 reserved hosts and TLDs, some of which are for development
 (https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml)

In this change, we update the host validator to:
 - Allow the `.dev` TLD
 - Disallow `example` hostname
 - Disallow the `.local`, `.example`, and `.test` reserved TLDs
 - Update all `example.com` references in specs to `notareal.domain`
 since `example.com` no longer validates

Let me know your thoughts on this!


app/models/host_validator.rb
```diff
@@ -19,6 +19,7 @@ def invalid?
   attr_reader :url
 
   def invalid_host?
+    puts host
```
gabebw: Looks like some debugging crept in 😄 

Can you remove this?

app/models/host_validator.rb
```diff
@@ -1,6 +1,6 @@
 class HostValidator
-  INVALID_HOSTS = %w(127.0.0.1 localhost)
-  INVALID_TLDS = %w(.dev)
+  INVALID_HOSTS = %w(127.0.0.1 localhost example.com example.org example.net example.dev)
```
gabebw: This sent me down a bit of a rabbithole - thank you!

I first visited https://example.com, which led me to https://www.iana.org/domains/reserved, which led me to [RFC 2606](https://tools.ietf.org/html/rfc2606) and [RFC 6761](https://tools.ietf.org/html/rfc6761), which describe example-type domains.

RFC 2606 specifically mentions:

* https://example.com
* https://example.net
* https://example.org

as reserved, all of which do indeed lead to IANA-maintained pages. 

It looks like `whois example.dev` shows that it's maintained by Google, `whois example.blog` is run by Automattic's registrar ("Knock Knock WHOIS there" 😄), and so it seems that indeed there's a pattern of `example.TLD` domains being registered.

Given that, what do you think about banning all `example.TLD` domains (using a method, not a giant blacklist array)? That's probably best done in a separate PR, so I suggest:

* leaving the `INVALID_HOSTS` variable alone in this PR
* opening a new PR (or I can do it) to ban all hosts with a domain of `example`.

spec/controllers/redirections_controller_spec.rb
```diff
@@ -16,7 +16,7 @@
     describe "GET #{action}" do
       it "creates a Redirection if it doesn't exist" do
         new_slug = "new"
-        url = "http://example.com"
+        url = "http://notareal.domain"
```
gabebw: Thanks for fixing the tests!

app/models/host_validator.rb
WilHall: > what do you think about banning all example.TLD domains (using a method, not a giant blacklist array)? That's probably best done in a separate PR

That sounds good to me! I'm happy to open that PR tomorrow morning.

gabebw: Thanks!

gabebw:
Looks like CircleCI has the "build PRs from forks" setting disabled. I'm going to turn on that setting for future PRs and just merge this one as-is.

spec/controllers/redirections_controller_spec.rb
gabebw: Oh wait, actually – could you revert these spec changes since they no longer are needed (I think)? I believe these changes should be folded into your future PR disallowing `example.com`.

spec/models/host_validator_spec.rb
```diff
@@ -4,7 +4,9 @@
   describe "#invalid?" do
     %w(
       http://localhost:3000
-      http://example.dev
+      http://devsite.example
```
gabebw: Please do keep this spec change, though.

PR #265, opened 2020-03-06
Use Ruby 2.7.0 (and Bundler 2)

gabebw:
It is the latest stable version of Ruby, and is officially supported by Heroku: https://devcenter.heroku.com/articles/ruby-support#supported-runtimes

Since the new CircleCI Docker image ships with Bundler 2 (specifically version 2.1.2), we now use Bundler 2 (check out the Gemfile.lock). At first, I put in a lot of work to keep using Bundler 1, but using Bundler 2 seems fine and easier.


gabebw:
Whew, okay. After some poking and prodding, this PR is ready for review.

.ruby-version
```diff
@@ -1 +1 @@
-2.6.5
+2.7.0
```
edwardloveall: Should just get rid of this file? Or get rid of `.tool-versions`? It feels weird to me having both.

gabebw: `.tool-versions` is just for asdf (my version manager). Since asdf can read `.ruby-version`, and so can everything else, we should standardize on `.ruby-version`. I'll remove `.tool-versions.

PR #271, opened 2020-03-12
Allow blacklisting certain URLs

### LINKED ISSUE #206: Allow blacklisting certain URLs
These URLs can't join the webring even if they try.

Idea: use a table named `blacklisted_referrers` with the following schema:

* `prefix` as `string`
* Rails standard timestamps

I suggest we check if an incoming URL *starts with* any of the blacklisted referrers. That way if we blacklist `evil.com`, we also blacklist `evil.com/~hello`, `evil.com/about`, but not `good.evil.com`.

I think we should not block a blacklisted URL's subdomains (at least for now) since usually subdomains are totally different sites (e.g. `something.tumblr.com` vs `another-thing.tumblr.com`).

There are some edge cases like `http://freepages.rootsweb.com/`, which uses directories, not subdomains, to indicate different sites: `http://freepages.rootsweb.com/~good` is totally different from `http://freepages.rootsweb.com/~bad`. However, this scheme also handles that because we can block `http://freepages.rootsweb.com/~bad`, which wouldn't touch `http://freepages.rootsweb.com/~good`.

gabebw:
The expected process for blacklisting:

* We manually add the host as a BlacklistedReferrer
* We manually unlink the offending Redirection

After that, whenever someone clicks "next" or "previous" from the offending site, they will be taken to the first Redirection and the site will _not_ be added to the ring.

Mostly fixes #206. Later, we'll probably want a specific page saying "you can't join" as detailed in #206. This PR just silently won't let them join.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,24 @@ Use `Redirection#unlink`:
 This will destroy the `Redirection` and re-link its ring neighbors, sealing the
 breach.
 
+## Blacklisting
+
+When blacklisting a URL, specify the bare URL, without `http` or `www`
+subdomain.
+
+To blacklist everything from `evil.com` but none of its subdomains:
+
+    BlacklistedReferrer.create!(host_with_path: "evil.com")
+
+To blacklist everything under the `/one/` directory (`evil.com/one/*`), but not
+the root (for example, to blacklist a user directory):
+
+    BlacklistedReferrer.create!(host_with_path: "good.com/~evil")
+
+To blacklist a subdomain:
+
+    BlacklistedReferrer.create!(host_with_path: "evil.good.com")
+
 ## Guidelines
 
 Use the following guides for getting things done, programming well, and
--- a/app/models/blacklist.rb
+++ b/app/models/blacklist.rb
@@ -1,0 +1,21 @@
+class Blacklist
+  def initialize(possible_match)
+    @possible_match = possible_match
+  end
+
+  def blacklisted?
+    if possible_match.nil?
+      false
+    else
+      host = URI.parse(possible_match).host
+
+      BlacklistedReferrer.where("host_with_path LIKE ?", "#{host}%").
+        or(BlacklistedReferrer.where("'www.'||host_with_path LIKE ?", "#{host}%")).
+        any?
+    end
+  end
+
+  private
+
+  attr_reader :possible_match
+end
--- a/db/migrate/20200312175018_create_blacklisted_referrers.rb
+++ b/db/migrate/20200312175018_create_blacklisted_referrers.rb
@@ -1,0 +1,10 @@
+class CreateBlacklistedReferrers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :blacklisted_referrers do |t|
+      t.string :host_with_path, null: false
+      t.index :host_with_path, unique: true
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -49,6 +49,20 @@ RSpec.describe RedirectionsController do
           end
         end
       end
+
+      context "when the URL is blacklisted" do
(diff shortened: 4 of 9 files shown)


app/models/blacklist.rb
```diff
@@ -0,0 +1,28 @@
...
+    if possible_match.nil?
+      false
+    else
+      uri = URI.parse(possible_match)
+      host = normalize_host(uri)
+      BlacklistedReferrer.all.any? do |referrer|
+        prefix_uri = URI.parse(referrer.prefix)
+
+        host == normalize_host(prefix_uri) &&
+          uri.path.starts_with?(prefix_uri.path)
```
edwardloveall: I'm finding this a little confusing. I think the part that's getting me is that we're storing the prefixes as one thing, and then normalize i when we take it out with `normalize_host`. Could we:

* Store as the normalized host in the first place
* Use a query to test this instead of loading all BlacklistedReferrers into memory?

I'm not really worried about optimization, but I think it could read better. For example if we store all urls without their protocol (which I assume we control because we'll be entering them) then the `else` block can turn into:

```ruby
host = URI.parse(possible_match).host
BlacklistedReferrer.where("prefix LIKE ?", "#{host}%").
  or(BlacklistedReferrer.where("prefix LIKE ?", "www.#{host}%")).
  any?
```

Downside is it involves SQL but I think it gets the point across more succinctly without compromising much readability. I'd love to know what you think!

db/migrate/20200312175018_create_blacklisted_referrers.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBlacklistedReferrers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :blacklisted_referrers do |t|
+      t.string :prefix, null: false
```
edwardloveall: I also wonder if we do the change I suggested above, it if makes more sense to do `host_with_path` or something to help denote what we want to store.

spec/controllers/redirections_controller_spec.rb
```diff
@@ -49,6 +49,20 @@
           end
         end
       end
+
+      context "when the URL is blacklisted" do
+        it "does not create a redirection" do
```
edwardloveall:
```suggestion
        it "does not create a redirection and redirects to the first redirection" do
```

app/models/blacklist.rb
gabebw: Nice! I like this a lot better. I'm definitely doing this.

db/migrate/20200312175018_create_blacklisted_referrers.rb
gabebw: Yep, good call.

README.md
```diff
@@ -33,6 +33,24 @@ Use `Redirection#unlink`:
 This will destroy the `Redirection` and re-link its ring neighbors, sealing the
 breach.
 
+## Blacklisting
+
+When blacklisting a URL, specify the bare URL, without `http` or `www`
```
gabebw: This is a bit confusing (it breaks if you blacklist `https://evil.com`), so I documented it. We could also build a `BlacklistCreationService` or something, but...that's a lot of ceremony for something we're going to do very rarely.

app/models/blacklist.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def blacklisted?
+    if possible_match.nil?
+      false
+    else
+      host = URI.parse(possible_match).host
+
+      BlacklistedReferrer.where("host_with_path LIKE ?", "#{host}%").
+        or(BlacklistedReferrer.where("'www.'||host_with_path LIKE ?", "#{host}%")).
```
edwardloveall: Oh huh, what does the `||` do? Concatenate?

gabebw:
Yep!

app/models/blacklist.rb
gabebw: Yep! I like it better than `SELECT CONCAT('www.', host_with_path)`.

PR #285, opened 2020-05-23
Disallow creating different slugs with similar URLs

gabebw:
When creating a new Redirection, this will normalize the URL by removing the protocol (HTTP/S) and any `www.` subdomain. So this:

```
https://www.foo.com/something?else=1
```

becomes this:

```
foo.com/something?else=1
```

Then we compare that normalized URL against the normalized URL of every other redirection, and mark the new record as invalid if there's a match.

-------------

This was prompted by a Redirection being added for `http://hotlinewebring.club`
when HLWR itself has `https://hotlinewebring.club`.

My big concern is:

* Running Ruby code every time we touch a Redirection is inefficient. Should we store the normalized URL in a field on the Redirection and re-calculate it whenever it's created/updated? Does it matter, given the number of records we're talking about?

### DIFF
--- a/spec/controllers/redirections_controller_spec.rb
+++ b/spec/controllers/redirections_controller_spec.rb
@@ -1,5 +1,19 @@
 require "rails_helper"
 
+RSpec.shared_examples_for "disallowing a similar URL" do |action, item|
+  it "does not add #{item[:new]} when #{item[:existing]} already exists" do
+    old_slug = "old"
+    new_slug = "new"
+    RedirectionCreation.perform(item[:existing], old_slug)
+
+    request.env["HTTP_REFERER"] = item[:new]
+    expect {
+      get action, params: { slug: new_slug }
+    }.to raise_error(ActiveRecord::RecordInvalid)
+    expect(Redirection.where(slug: new_slug)).to be_empty
+  end
+end
+
 RSpec.describe RedirectionsController do
   describe "GET :next" do
     it "redirects to the next site" do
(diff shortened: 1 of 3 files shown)


edwardloveall:
I don't think we should worry about efficiency at this point. If you were really worried, you could perhaps do a single check in SQL first but I don't think the complexity is worth it at this point in HLWR's life.

gabebw:
>Since this logic is in the model, I'm surprised you put the tests in the controller. Seems like it might be simpler to check the validation there.

Aha, yes, I meant to ask you about that! Thank you.

I started with controller tests because I wanted a feature(ish) test to make sure that it would actually stop these kinds of duplicates in a real-world scenario.

What do you think about keeping just two tests in the controller (1 for exact match and 1 for a similar dupe), and moving the rest to the model?

edwardloveall:
>What do you think about keeping just two tests in the controller (1 for exact match and 1 for a similar dupe), and moving the rest to the model?

Works for me!

spec/controllers/redirections_controller_spec.rb
```diff
@@ -1,5 +1,19 @@
 require "rails_helper"
 
+RSpec.shared_examples_for "disallowing a similar URL" do |action, item|
```
edwardloveall, line 3: How would you feel about putting this logic inline in each test? It's only two tests, and I think it's easier to understand what's required to set up the test. If you still feel like the length is hurting readability, maybe just a method would be a good middle-ground instead of this RSpec-only feature?

I've never used shared examples and it took me some time to understand what was happening. That frustration is surely coloring my opinion here. See also [my recent blog post](https://thoughtbot.com/blog/the-self-contained-test) on keeping specs all in one place.

gabebw: Absolutely. Now that there are fewer controller tests, I'll remove this block.

PR #286, opened 2020-06-06
Make the link audit task more resilient

gabebw:
* Just like before, if there's a `SocketError`, indicate that the website is completely offline
* If there is any other kind of error (`StandardError`), print the error message

The certificate for https://irongeek.net expired and the `rake link_audit` task crashed when it got to it. Now it prints the error and keeps going.

To accommodate showing more information, the `MissingLinkFinder` now sends more information back to the link audit task, including a `status` plus information relating to that status.

Alacritty 2020-06-05 20-15-09

### DIFF
--- a/app/models/missing_link_finder.rb
+++ b/app/models/missing_link_finder.rb
@@ -10,8 +10,30 @@ class MissingLinkFinder
   end
 
   def run
-    html = fetch(redirection.url)
+    begin
+      response = fetch(redirection.url)
+      if response.is_a?(Net::HTTPSuccess)
+        missing = missing_links(response.body)
+        if missing.empty?
+          { status: :good }
+        else
+          { status: :missing_links, missing: missing }
+        end
+      else
+        { status: :offline }
+      end
+    rescue SocketError
+      { status: :offline }
+    rescue StandardError => error
+      { status: :error, error: "#{error.class}: #{error.message}" }
+    end
+  end
 
+  private
+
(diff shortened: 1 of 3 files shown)


app/models/missing_link_finder.rb
```diff
@@ -10,8 +10,28 @@ def initialize(redirection)
   end
 
   def run
-    html = fetch(redirection.url)
+    begin
+      missing = missing_links
```
gabebw, line 14: It's kind of gross that `missing` can be `nil` or an empty array or a non-empty array.

app/models/missing_link_finder.rb
```diff
@@ -10,8 +10,30 @@ def initialize(redirection)
   end
 
   def run
-    html = fetch(redirection.url)
+    begin
+      html = fetch(redirection.url)
+      if html.nil?
```
gabebw, line 15: This means the status wasn't successful (4xx or 5xx response code).

edwardloveall: Would it make sense to actually make the status part of the conditional? Or maybe we could move to a response type or something? `html.nil?` feels very implicit, but maybe I'm over thinking it.

gabebw: >Would it make sense to actually make the status part of the conditional?

Oh yeah, great idea. `fetch` is hiding too much here. I'll open it up.

PR #287, opened 2020-06-06
Use newer Spring binstubs to avoid warning

gabebw:
`bin/spring` was using `Gem.paths=` which printed this warning whenever `bin/rspec` (which uses Spring) ran:

>Array values in the parameter to `Gem.paths=` are deprecated.

Now it uses different code and does not print a warning.

### DIFF
--- a/bin/spring
+++ b/bin/spring
@@ -1,15 +1,17 @@
 #!/usr/bin/env ruby
 
-# This file loads spring without using Bundler, in order to be fast.
+# This file loads Spring without using Bundler, in order to be fast.
 # It gets overwritten when you run the `spring binstub` command.
 
 unless defined?(Spring)
-  require "rubygems"
-  require "bundler"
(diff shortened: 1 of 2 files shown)


bin/spring
```diff
@@ -1,15 +1,17 @@
...
 
-# This file loads spring without using Bundler, in order to be fast.
+# This file loads Spring without using Bundler, in order to be fast.
 # It gets overwritten when you run the `spring binstub` command.
 
 unless defined?(Spring)
-  require "rubygems"
-  require "bundler"
+  require 'rubygems'
+  require 'bundler'
```
edwardloveall, line 8: Do you have a linter doing some stuff? Or are you a... 😎 single man, now?

gabebw: This is 100% Spring: I ran `spring binstub rspec` and this is what it generated.

PR #296, opened 2020-08-13
Improve link audit, including iframe support

### LINKED ISSUE #180: Support frames in the `link_audit` rake task
If you visit http://bitfission.com, you can see that the links show up just fine. However, the HTML shows this:

```html
<FRAMESET COLS="222px,*">
  <FRAME SRC="side.html" NAME=side>
  <FRAME SRC="about.html" NAME=main>
</FRAMESET>
```

The HLWR links are in `about.html`.

I think the robust way to solve this is to use something like Capybara and/or Chromedriver to see what a user would actually see.

We'd need to add a buildpack (https://elements.heroku.com/buildpacks/heroku/heroku-buildpack-google-chrome) to Heroku to support running it in production:

```
heroku buildpacks:add heroku/google-chrome --remote [whatever]
```

gabebw:
* `MissingLinkFinder` learned the difference between a 404 (`:not_found`) and an offline (`:offline`) site, and we show more informative messages as a result
* Capybara Mechanize, which is like Rack Test (no JS support) but runs against external websites, lets us scan `<iframe>` and `<frame>` tags

Fixes #180.


app/models/missing_link_finder.rb
```diff
@@ -1,3 +1,11 @@
+require "capybara/mechanize"
```
gabebw, line 1: I recommend viewing this file in side-by-side diff mode, since it changed significantly. It used to use simple `Net::HTTP` requests but now spins up a (non-Javascript) Capybara session, which lets us easily find elements like the iframes and frames (and next/prev links), and adds a useful concept of "the current page".

spec/models/missing_link_finder_spec.rb
```diff
@@ -18,6 +18,25 @@
       expect(result).to eq({ status: :good })
     end
 
+    it "follows redirects" do
```
gabebw, line 21: I figured I might as well add this test.

app/models/missing_link_finder.rb
edwardloveall: Hah, I was having trouble reading it, then I did this very thing. :high-five:

app/models/missing_link_finder.rb
```diff
@@ -7,40 +15,63 @@ class MissingLinkFinder
...
 
-  def missing_links(html)
-    if html
-      missing_links = []
-      missing_links << "next" if !html.match?(next_link)
-      missing_links << "prev" if !html.match?(prev_link)
+  def missing_links
+    found_next = has_next_link?
+    found_previous = has_prev_link?
+    if !found_next || !found_previous
```
edwardloveall, line 51: I'm finding the negation and or a little weird to read (not this or not that). I wonder if instead of `has_<some>_link` the methods were `missing_<some>_link?` instead? Alternatively, we could return early:

```suggestion
  return if has_next_link? && has_prev_link?
```

app/models/missing_link_finder.rb
```diff
@@ -7,40 +15,63 @@ class MissingLinkFinder
...
+      session.all("iframe, frame").map { |i| i[:src] }.each do |iframe_src|
+        unless iframe_src.start_with?("http")
```
edwardloveall, line 53: Am I reading this correctly as: "**Don't** parse the iframe's `src` attribute into a URI if it starts with `http`"?

Why would we not want to parse it if it started with a protocol?

app/models/missing_link_finder.rb
```diff
@@ -7,40 +15,63 @@ class MissingLinkFinder
...
+          uri = URI.parse(session.current_url)
+          iframe_src = "#{uri.scheme}://#{uri.host}/#{iframe_src}"
+        end
 
-      missing_links
+        session.visit iframe_src
+        found_next ||= has_next_link?
+        found_previous ||= has_prev_link?
```
edwardloveall, line 60: Just want to talk this one through:

* `found_<link>` is true
  * doesn't change `found_<link>` because the right side of the `||=` condition wouldn't trigger
* `found_<link>` is false because it couldn't find a link earlier
  * it executes `has_<some>_link` because falsey values will execute the right side of the `||=`

spec/models/missing_link_finder_spec.rb
```diff
@@ -33,7 +52,57 @@
...
+        <a href="https://hotlinewebring.club/#{redirection.slug}/next">next</a>
+      IFRAME
+      body = <<~BODY
+        <iframe src="iframe_relative_previous">
+        <iframe src="#{redirection.url}/iframe_next">
+      BODY
+
+      stub_request(:get, redirection.url).to_return(body: body)
+      stub_request(:get, "#{redirection.url}/iframe_relative_previous")
+        .to_return(body: iframe_relative_previous)
```
edwardloveall, line 74: Nice.

app/models/missing_link_finder.rb
gabebw: Yeah, this is confusing. I regret >50% of the `unless`es I write and this certainly counts 😄 I'll fix it, but here's the background:

The problem is that some iframe `src`es are relative, not absolute URLs. For example, on bitfission.com, we have this:

```html
<frame src="title.html" name="TITLE" scrolling="NO">
```

Capybara crashes when it tries to `visit` a relative URL, so we need to turn `title.html` into `http://bitfission.com/title.html`, which is what this block is doing. The URI parsing is not to get a URI per se; it's to parse out `http://bitfission.com` so we can add it to the relative `src`.

app/models/missing_link_finder.rb
gabebw: Yep, exactly. If we've already found a link, we can't **un**find it, so this ensures that if we're searching a page with 3 `<frame>`s and we found a link on the first frame but not the others, it still counts as found rather than it needing to be found on the exact same page every time.

gabebw:
@edwardloveall I've addressed your comments in 87b30ada02e06ceea9e2bac62a87065dd843b86b.

I moved the check for `if missing_next || missing_previous` into the `session.all("iframe, frame")` block as well, so that we stop looking when we find both links. It's not an optimization that I can imagine ever mattering in practice (unless someone has a page with thousands of iframes and the links are in the first one?), but I think it communicates intent better.

app/models/missing_link_finder.rb
```diff
@@ -7,58 +15,78 @@ class MissingLinkFinder
...
-      missing_links = []
-      missing_links << "next" if !html.match?(next_link)
-      missing_links << "prev" if !html.match?(prev_link)
+  def missing_links
+    missing_next = !has_next_link?
+    missing_previous = !has_prev_link?
+    session.all("iframe, frame").map { |i| i[:src] }.each do |iframe_src|
+      if missing_next || missing_previous
+        session.visit ensure_absolute_path(iframe_src)
+        if has_next_link?
```
gabebw, line 54: I elected to use `if` statements instead of the following code, which I think is some of the most confusing I've seen in years 😄 

```ruby
missing_next &&= !has_next_link?
missing_previous &&= !has_prev_link?
```

edwardloveall: Wow. I did not know that was even valid ruby. Good call.

edwardloveall:
This all looks great!

PR #298, opened 2020-08-13
Make capybara-mechanize available in production

gabebw:
It broke the whole website (now rolled back) because we try to load capybara-mechanize when we preload all the classes.

I tested this branch on staging, so it definitely works now.


edwardloveall:
Sorry this took me so long. I forgot about it.

gabebw:
@edwardloveall 💖 Thank you for the review!

PR #301, opened 2020-08-29
Check for both next/prev links

gabebw:
And write some dang tests!

### DIFF
--- a/app/models/missing_link_finder.rb
+++ b/app/models/missing_link_finder.rb
@@ -103,7 +103,7 @@ class MissingLinkFinder
       # all `<a>` tags inside an `<svg>` and iterate through them.
 
       session.all("svg a").any? do |anchor|
-        anchor["xlink:href"].match?(next_link)
+        anchor["xlink:href"].match?(link)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/models/missing_link_finder.rb
```diff
@@ -103,7 +103,7 @@ def has_link?(link)
       # all `<a>` tags inside an `<svg>` and iterate through them.
 
       session.all("svg a").any? do |anchor|
-        anchor["xlink:href"].match?(next_link)
```
gabebw, line 106: We were only checking for the "next" link before.

spec/models/missing_link_finder_spec.rb
```diff
@@ -174,5 +176,38 @@
...
+          body = <<~BODY
+            <svg>
+              <a xlink:href="#{link}">#{type}</a>
+            </svg>
+          BODY
+          stub_request(:get, redirection.url).to_return(body: body)
+
+          result = MissingLinkFinder.new(redirection).run
+
+          missing = %w[next prev] - [type]
```
edwardloveall, line 207: At first I thought iterating over the types to make two tests easy would be fine, but this and the `link` variable above are starting to push into too much logic in the test itself I think. Up to you though if you want to rewrite it.

gabebw: I agree, it's a bit complex. I split it into two tests: 09006574e8710135e866f1ec747e3ca2a2c2d301

PR #302, opened 2020-09-11
[Security] Update actionview to >= 6.0.3.3

gabebw:
Security advisory: https://github.com/advisories/GHSA-cfjv-5498-mph5

The advisory does not apply to our code, but why not keep up Hotline Webring's stellar record of updates?


edwardloveall:
That's a weird test failure. I'm not really sure what it's trying to say. It almost seems like the error is that we updated the Gemfile and the fix is to update the Gemfile.lock, but the diff is all Gemfile.lock

gabebw:
I changed the `Gemfile`, then ran `bundle install`, then changed the Gemfile again, then DID NOT run `bundle install`. My bad. Just pushed up a commit that should fix it: b1ef97ab16eea94cbae229263079fa56a3f1bfad

edwardloveall (approved):

PR #303, opened 2020-09-23
Log headers on redirection failure

edwardloveall:
We've had a few times where we wanted to see some evidence of redirection creation or failure, but heroku doesn't log headers (like the HTTP referrer).

I've added logging for when we do or do not have a referrer so we have a little more idea about what might have gone wrong in the redirection creation process.

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -19,11 +19,23 @@ class RedirectionsController < ApplicationController
   def find_or_create_redirection
     Redirection.find_by(slug: params[:slug]) ||
       RedirectionCreation.perform(referrer, params[:slug]) ||
-      Redirection.first
+      log_headers_and_use_fallback_redirection
+  end
+
+  def log_headers_and_use_fallback_redirection
+    logger.info "Redirection creation failed. Here are the headers:"
+    logger.info http_headers
+    Redirection.first
   end
 
   def referrer
     request.env["HTTP_REFERER"]
+  end
+
+  def http_headers
+    http_only = ->(key, _) { key.start_with?("HTTP") }
+    cookie_header = ->(key, _) { key == "HTTP_COOKIE" }
+    request.env.select(&http_only).reject(&cookie_header)
   end
 
   def ensure_referrer_is_not_localhost


gabebw:
What do you think about passing in all of the HTTP headers (`request.env`?) and logging them when the referrer isn't present? That way we can see what IS there and what isn't, and maybe write new code to handle new cases.

edwardloveall:
Works for me! I have to figure out if I want to pass in all the headers into the creation object or refactor the controller to fail, log, and recover instead of recovering immediately.

edwardloveall:
@gabebw done. I actually misread what you said a bit and did a slightly different thing of:

* if we can't find a redirection
* and we can't create it
* log all the headers and return our fallback redirection (`Redirection.first`)

I think this does effectively the same thing and it the `RedirectionCreation` object from having to accept _all_ headers. Lemme know what you think.

gabebw (approved):

gabebw:
I like your solution better than what I suggested. This way, `RedirectionCreation` stays simple (and easy to test), and `request.env` stays in the controller where it belongs.

app/controllers/redirections_controller.rb
```diff
@@ -19,7 +19,13 @@ def previous
   def find_or_create_redirection
     Redirection.find_by(slug: params[:slug]) ||
       RedirectionCreation.perform(referrer, params[:slug]) ||
-      Redirection.first
+      log_headers_and_use_fallback_redirection
+  end
+
+  def log_headers_and_use_fallback_redirection
+    logger.info "Redirection creation failed. Here are the headers:"
+    logger.info request.env
```
gabebw, line 27: This has a lot of stuff that isn't just `headers` - can we trim it down to just the headers?

gabebw (changes requested):

app/controllers/redirections_controller.rb
gabebw: Screenshot of everything that gets logged:

hotline-webring-production - Papertrail 2020-09-26 12-06-42

gabebw: One option, though not totally foolproof, is to only select keys that are all uppercase:

```ruby
headers = request.env.select { |(key, _)| key.upcase == key }
```

gabebw: Ah, there is a foolproof method, from [the Rack spec](https://github.com/rack/rack/blob/master/SPEC.rdoc):

>`HTTP_` Variables
>Variables corresponding to the client-supplied HTTP request headers (i.e., variables whose names begin with `HTTP_`). The presence or absence of these variables should correspond with the presence or absence of the appropriate HTTP header in the request. See RFC3875 section 4.1.18 for specific behavior.

So we can do this:


```ruby
headers = request.env.select { |(key, _)| key.start_with?("HTTP_") }
```

gabebw: And to cut down on noise, we might want to remove `HTTP_COOKIE`, which is a long string like this:

```
"HTTP_COOKIE"=>"_hotline-webring_session=d3M5MzQzdVhHeW5uSnVaWmNNdXpLYUg5a0NnWGFJWWdKckRNWXlGN2hTbTB5TTBIVnExcUhHcGJvaU03ZHBySitndnY0V3BlWnYrSTJlTGNxU3g3WUt1MEgvcWVoVWl0aGt5NGRqQmN4UVJMYWZGZGNqSXVXUEx4UXh0UUpyT0lWVFpqbDF2OThmWnhvQldQMnZpeWJ3PT0tLWwwMVZMeCtyeTk5WVJXNjk4NWV6aVE9PQ%3D%3D--1851b3c61a861f212c969ddaa493b5dd21471abb"
```

```ruby
headers = request.env.select { |(key, _)| key.start_with?("HTTP_") && key != "HTTP_COOKIE" }
```

edwardloveall: I also found https://stackoverflow.com/a/51884276/638966 which mentions that effectively no HTTP header has a `.` in it. This does show us a few other headers but nothing that I think we'll care about or will be different from request to request. I like your HTTP but not COOKIE idea.

edwardloveall: Boom dc93b2b

gabebw (approved):

PR #310, opened 2021-03-18
Ignore trailing slash when comparing URLs

gabebw:
Someone created a Redirection for "hotlinewebring.club/" with a trailing slash. Let's prevent that in the future.


edwardloveall (approved):
🎉

PR #311, opened 2021-03-18
Add logging when redirection creation fails

gabebw:
Tagging it with the slug will make it easy to search in Papertrail.

Sample output from running it locally:

    [asdfasdf] Redirection creation failed.
    [asdfasdf] Referrer is BLANK!
    [asdfasdf] Here are all of the HTTP headers:
    [asdfasdf] {"HTTP_VERSION"=>"HTTP/1.1", "HTTP_HOST"=>"localhost:3000", "HTTP_USER_AGENT"=>"curl/7.64.1", "HTTP_ACCEPT"=>"*/*"}

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -23,8 +23,17 @@ class RedirectionsController < ApplicationController
   end
 
   def log_headers_and_use_fallback_redirection
-    logger.info "Redirection creation failed. Here are the headers:"
-    logger.info http_headers
+    tagged_log "Redirection creation failed."
+    if request.env["HTTP_UPGRADE_INSECURE_REQUESTS"].present?
+      tagged_log "It likely failed because the links use HTTP instead of HTTPS"
+    end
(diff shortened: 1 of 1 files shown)


app/controllers/redirections_controller.rb
```diff
@@ -23,8 +23,17 @@ def find_or_create_redirection
   end
 
   def log_headers_and_use_fallback_redirection
-    logger.info "Redirection creation failed. Here are the headers:"
-    logger.info http_headers
+    tagged_log "Redirection creation failed."
+    if request.env["HTTP_UPGRADE_INSECURE_REQUESTS"].present?
+      tagged_log "It likely failed because the links use HTTP instead of HTTPS"
```
gabebw, line 28: This is what happened here: https://my.papertrailapp.com/events?focus=1309069283861745671&selected=1309069283861745671

Now it's easy to see that this was the cause.

edwardloveall (approved):

PR #314, opened 2021-03-20
Automatically deploy to Heroku on merge

### LINKED ISSUE #313: We should set up continuous auto-deploys to Heroku staging/production
Whenever something is merged to `main`, deploy it to staging and if that works, deploy to production.

gabebw:
Closes #313

### DIFF
--- a/.github/workflows/heroku-deploy.yml
+++ b/.github/workflows/heroku-deploy.yml
@@ -1,0 +1,34 @@
+name: Heroku Deploy
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - main
+# `types: [closed]` above means that this step runs even when a PR is
+# closed without merging.
+# This `if` statement ensures we only run this step when the PR is merged.
+if: ${{ github.event.pull_request.merged == true }}
+
+jobs:
+  heroku-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


gabebw:
I confirm that this worked to deploy to staging and correctly did not deploy to production: https://github.com/hotline-webring/hotline-webring/runs/2156227279?check_suite_focus=true

Now I'm going to change the syntax so it only runs when we want, and not on every push. I'll squash all of the commits before merging.

.github/workflows/heroku-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - main
+# `types: [closed]` above means that this step runs even when a PR is
+# closed without merging.
+# This `if` statement ensures we only run this step when the PR is merged.
+if: ${{ github.event.pull_request.merged == true }}
```
edwardloveall, line 11: So this all together means that we only deploy when we close (merge?) a PR into main?

edwardloveall (approved):
One question, but if it's what I think it means, then it looks good to me. Thanks for doing this!

.github/workflows/heroku-deploy.yml
gabebw: Yes, you're exactly right.

> close (merge?)

Yeah, there's no `merge` event, so we have to check for when PRs are closed (which means closed or merged) then confirm that it's a merge.

PR #315, opened 2021-03-21
Fix workflow syntax

gabebw:
`if` should be at the job level, not the top level.

The most recent workflow failed on merge: https://github.com/hotline-webring/hotline-webring/actions/runs/673952019


edwardloveall (approved):
We'll do it live.

PR #318, opened 2021-03-21
Correctly read the base ref for the PR

gabebw:


gabebw:
This worked! https://github.com/hotline-webring/hotline-webring/runs/2161568715?check_suite_focus=true

PR #320, opened 2021-03-30
Block IPs in the range 192.168.*.*

### LINKED ISSUE #319: Block 192.168.*.* from joining
This is represented in Ruby like this:

```ruby
IPAddr.new('192.168.0.0/16')
```

Then we can check:

```
> IPAddr.new('192.168.0.0/16') === '192.168.1.220'
true
```

The block goes in the `HostValidator` class.

gabebw:
This also more-correctly blocks the IPs that we already blocked.

Fixes #319.

### DIFF
--- a/app/models/host_validator.rb
+++ b/app/models/host_validator.rb
@@ -1,6 +1,12 @@
 class HostValidator
-  INVALID_HOSTS = %w(0.0.0.0 127.0.0.1 localhost)
+  INVALID_HOSTS = %w(localhost)
   INVALID_TLDS = %w(.example .local .test)
+  # https://en.wikipedia.org/wiki/Reserved_IP_addresses
+  INVALID_IP_RANGES = %w(
+    0.0.0.0/8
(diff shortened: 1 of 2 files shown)


app/models/host_validator.rb
```diff
@@ -26,7 +32,24 @@ def invalid_tld?
...
+
   def host
     URI.parse(url).host
   end
+
+  def is_ip?(host)
+    begin
+      IPAddr.new(host)
+      true
+    rescue IPAddr::InvalidAddressError
```
gabebw, line 51: `IPAddr.new` throws this error if it's an invalid IP. We could try matching by using internal regex constants like `IPAddr::RE_IPV4ADDRLIKE` / `IPAddr::RE_IPV6ADDRLIKE_COMPRESSED` / `IPAddr::RE_IPV6ADDRLIKE_FULL` but I'm OK with this as-is.

edwardloveall (approved):
Awesome, thank you!

PR #323, opened 2021-05-18
Restyle without bourbon/neat

edwardloveall:
We were not really using bourbon/neat anyway in the styles and I totally over-engineered them when we first created the site. This dramatically simplifies them and removes our dependency on the ruby sass library which is deprecated.

There are some minor differences that I didn't bother to get pixel perfect because I didn't think it mattered. Let me know if you want me to get it more exact!

## Before

prod

## After

dev


gabebw (approved):
Brilliant. Thank you!

A `+230 −556` PR is always great.

PR #331, opened 2021-08-24
Correctly match blocked referrer paths

gabebw:
Our previous code used simple string comparison, which meant that if `evil.com/subpath` was blocked, we would also block everything under `evil.com/subpathABC` (since it starts with `evil.com/subpathABC`).

The fix is to use code that understands that the URLs are actually URLs, not simple strings, so that we can correctly compare the hosts and paths.

This uses `Pathname#ascend` (docs: https://ruby-doc.org/stdlib-2.5.3/libdoc/pathname/rdoc/Pathname.html#method-i-ascend) which yields an iterator over the parent directories of a given Pathname.

For example, given `/foo/bar/baz`, `ascend` yields `/foo/bar/baz`, then `/foo/bar`, then `/foo`, then `/`. The yielded values are Pathnames, not strings.

### DIFF
--- a/app/models/block.rb
+++ b/app/models/block.rb
@@ -7,16 +7,28 @@ class Block
     if possible_match.nil?
       false
     else
-      uri = URI.parse(possible_match)
-      host_with_path = "#{uri.host}#{uri.path}"
+      uri = URI.parse(possible_match.downcase)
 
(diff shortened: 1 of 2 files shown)


edwardloveall (approved):
Didn't know about `Pathame#ascend`. Pretty cool!

PR #336, opened 2021-12-17
Version all old ActiveRecord migration classes

gabebw:
Rails 7 will throw an error about unversioned migrations:

    Directly inheriting from ActiveRecord::Migration is not supported.
    Please specify the Active Record release the migration was written
    for.


gabebw:
I love how minimal this project is. We should be using fewer railties, too 😄

gabebw:
>We should be using fewer railties, too 😄

In fact: #337

PR #337, opened 2021-12-18
Remove everything about ActionMailer and ActiveJob

gabebw:
We can't remove the gems since `rails` pulls them in, but we can avoid loading them in the first place.


edwardloveall (approved):

edwardloveall:
Looks like we need to also remove `spec/support/action_mailer.rb`

PR #340, opened 2021-12-20
Upgrade to Rails 7

gabebw:
* Load config defaults for 7.0
* Accept the double quotes that Rails now prefers
* Accept `require_relative` instead of the more-verbose `File.expand_path` of yore
* Raise errors on disallowed deprecations (but don't disallow anything yet, just like the default generated Rails settings)

Accept a change:

* Stop always eager-loading the app in tests (`config.eager_load`)


config/application.rb
```diff
@@ -10,7 +10,7 @@
 
 module HotlineWebring
   class Application < Rails::Application
-    config.load_defaults 6.1
+    config.load_defaults 7.0
```
gabebw, line 13: Here's what this does: https://guides.rubyonrails.org/configuring.html#results-of-config-load-defaults

Honestly, I don't know what every setting does, but we do so little in this app that if the tests pass, I'm confident that we're fine. We could break cookies and it would be fine.

config/environments/development.rb
```diff
@@ -1,28 +1,43 @@
+require "active_support/core_ext/integer/time"
```
gabebw, line 1: I think this is for `2.days` in the `Cache-Control` header.

config/environments/development.rb
```diff
@@ -1,28 +1,41 @@
...
+
 Rails.application.configure do
   config.cache_classes = false
   config.eager_load = false
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local = true
+
+  # Tell Rails not to send server timing headers
+  config.server_timing = false
```
gabebw, line 9: Server Timing headers provide a header with how long the view, SQL, etc each took.

From the [7.0.0 release notes](https://github.com/rails/rails/releases/tag/v7.0.0):

>Add Server-Timing middleware
>Server-Timing specification defines how the server can communicate to browsers performance metrics
about the request it is responding to.
> The ServerTiming middleware is enabled by default on development environment by default using the
config.server_timing setting and set the relevant duration metrics in the Server-Timing header
> The full specification for Server-Timing header can be found in: https://www.w3.org/TR/server-timing/#dfn-server-timing-header-field

I disabled it because:

1) We definitely don't need these kinds of metrics, even in development
2) It tells Rails to do more work, so why not tell it to do less?

config/environments/development.rb
```diff
@@ -1,28 +1,41 @@
...
   end
 
   config.active_support.deprecation = :raise
+  # Raise exceptions for disallowed deprecations.
+  config.active_support.disallowed_deprecation = :raise
+  # Tell Active Support which deprecation messages to disallow.
+  config.active_support.disallowed_deprecation_warnings = []
+
   config.active_record.migration_error = :page_load
+  config.active_record.verbose_query_logs = true
```
gabebw, line 33: This was `true` before (see line 27 below) but I put all the `config.active_record` lines together.

config/environments/development.rb
```diff
@@ -1,28 +1,41 @@
...
+
   config.assets.debug = true
   config.assets.digest = true
+  config.assets.quiet = true
```
gabebw, line 37: This was `true` before (see line 26 below) but I put all the `config.assets` lines together.

app/controllers/redirections_controller.rb
```diff
@@ -5,13 +5,13 @@ class RedirectionsController < ApplicationController
   def next
     redirection = find_or_create_redirection
 
-    redirect_to redirection.next_url
+    redirect_to redirection.next_url, allow_other_host: true
```
edwardloveall, line 8: Oh, neat!

edwardloveall (approved):
_TV commercial voice_: Give the double quotes that growing Rails' craves

PR #342, opened 2021-12-24
Remove Spring

gabebw:
It's no longer turned on in Rails 7 by default:

> Spring is no longer on by default, as faster computers have made it less relevant on anything but the largest applications.

(Source: the Rails 7 announcement at https://rubyonrails.org/2021/12/15/Rails-7-fulfilling-a-vision)


Gemfile
```diff
@@ -13,8 +13,6 @@ gem "rails", "~> 7.0.0"
 gem "sassc-rails"
 
 group :development do
-  gem "spring"
```
gabebw, line 16: It's still in the `Gemfile.lock` since the `rails` gem depends on it, but we no longer load it.

edwardloveall (approved):
Update Gemfile.lock?

gabebw:
#343 is the followup

PR #346, opened 2022-02-05
Add an admin panel

gabebw:
This will let us make quick changes without opening up the Heroku console.

It is responsive, so we can also do it on our phones.

To see it, go to https://hotline-webring-staging.herokuapp.com/admin and use the username/password that I will tell you.

## Full Size

CleanShot 2022-02-04 at 18 07 42@2x

## Mobile

CleanShot 2022-02-04 at 18 08 06@2x

## Screenshot from my actual phone after Blocking & Unlinking

IMG_3929

### DIFF
--- a/.sample.env
+++ b/.sample.env
@@ -1,9 +1,11 @@
 # http://ddollar.github.com/foreman/
-ASSET_HOST=localhost:3000
+ADMIN_PASSWORD=admin
+ADMIN_USERNAME=admin
 APPLICATION_HOST=localhost:3000
+ASSET_HOST=localhost:3000
+EXECJS_RUNTIME=Node
 RACK_ENV=development
 SECRET_KEY_BASE=development_secret
-EXECJS_RUNTIME=Node
 SMTP_ADDRESS=smtp.example.com
 SMTP_DOMAIN=example.com
 SMTP_PASSWORD=password
--- a/app/assets/stylesheets/_flashes.scss
+++ b/app/assets/stylesheets/_flashes.scss
@@ -1,0 +1,26 @@
+.flashes {
+  .flash-alert,
+  .flash-error,
+  .flash-notice,
+  .flash-success {
+    border-radius: 4px;
+    margin-bottom: 1rem;
+    padding: 1rem;
+  }
+
+  .flash-alert {
+    background: $pastel-yellow;
+  }
+
+  .flash-error {
+    background: $pastel-red;
+  }
+
+  .flash-notice {
+    background: $pastel-blue;
+  }
+
+  .flash-success {
+    background: $pastel-green;
+  }
+}
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -1,0 +1,104 @@
+@charset "utf-8";
+
+@import "normalize-rails";
+@import "variables";
+@import "flashes";
+
+header {
+  display: flex;
+  background: $hotline-bling-pink;
+  padding: 0 1em;
+  margin-bottom: 2em;
+}
+
+main {
+  // Make it larger so it looks nice on a phone
+  font-size: 18px;
+  margin: 0 auto;
+  max-width: $admin-max-width;
+}
+
+.redirection {
+  border-bottom: 1px solid black;
+  display: flex;
+  flex-direction: column;
+  justify-items: center;
+  margin-bottom: 1em;
+
+  > div {
+    margin-bottom: 1rem;
+    display: grid;
+    grid-template-columns: 1fr 3fr;
+  }
+}
+
+.url,
+.original-url {
+  word-wrap: break-word;
+  word-break: break-all;
+}
+
+label {
+  font-weight: bold;
+}
+
+.content {
+}
+
+.block-button button {
+  -webkit-font-smoothing: antialiased;
+  background-color: $dark-pink;
+  border-radius: 4px;
+  border: 0;
+  color: white;
+  padding: 0.5em 0.75em;
+  transition: background-color 0.3s ease-in;
+  line-height: 1;
+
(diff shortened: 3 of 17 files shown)


app/assets/stylesheets/_flashes.scss
```diff
@@ -0,0 +1,30 @@
...
+  .flash-notice,
+  .flash-success {
+    border-radius: 4px;
+    margin-bottom: 1rem;
+    padding: 1rem;
+  }
+
+  .flash-alert {
+    // yellow
+    background: #fff3cd;
```
edwardloveall, line 13: What do you think about putting these colors in `app/assets/stylesheets/_variables.scss`?

app/assets/stylesheets/admin.scss
```diff
@@ -0,0 +1,103 @@
...
+  display: flex;
+  background: $hotline-bling-pink;
+  padding: 0 1em;
+  margin-bottom: 2em;
+}
+
+main {
+  font-size: 18px;
+  margin: 0 auto;
+  max-width: 80%;
```
edwardloveall, line 17: What about these as variables, also? I'm thinking about all these mystery guests and wanting to refactor the CSS at some point soon. I'm worried that trying to untangle all of the unlabeled values will be a pain 😕

app/models/blocked_referrer.rb
```diff
@@ -1,2 +1,17 @@
 class BlockedReferrer < ActiveRecord::Base
+  # Uses the URL's host, unless it's a sites.google.com link, in which case we
+  # include the site name: "sites.google.com/sites/foobar".
```
edwardloveall, line 3: This took me a second to figure out that what we're doing is removing the trailing path info past `/sites/foobar/`. At least I think that's what we're doing. Is that right?

If so, would this be a better description?

```suggestion
  # Uses the URL's host, unless it's a sites.google.com link, in which case we
  # only include the site's "root" path:
  # sites.google.com/sites/foobar/baz/bat -> sites.google.com/sites/foobar
```

app/assets/stylesheets/admin.scss
```diff
@@ -0,0 +1,103 @@
...
+
+.redirection {
+  border-bottom: 1px solid black;
+  display: flex;
+  flex-direction: column;
+  justify-items: center;
+  margin-bottom: 1em;
+}
+
+.chunk {
```
edwardloveall, line 28: Not a big deal, but would it be nicer to avoid all the `chunk` class declarations and redefine this as:

```suggestion
.redirection > div {
```

spec/features/admin_blocks_a_redirection_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+RSpec.feature "Admin blocks and unlinks a redirection" do
+  before do
+    page.driver.browser.basic_authorize(correct_username, correct_password)
+  end
```
edwardloveall, line 6: What do you think about making a small helper method and removing the need for `before` altogether?

```rb
def login
  page.driver.browser.basic_authorize(
    ENV.fetch("ADMIN_USERNAME"),
    ENV.fetch("ADMIN_PASSWORD")
  )
end
```

And then of course in each test:

```rb
scenario "blocks and unlinks a redirection" do
  redirection = create(...)

  login
  visit admin_root_path
```

spec/models/block_spec.rb
```diff
@@ -57,6 +57,27 @@
       end
     end
 
+    context 'blocking www.evil.com' do
```
edwardloveall, line 60:
```suggestion
    context "blocking www.evil.com" do
```

spec/models/blocked_referrer_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe BlockedReferrer do
+  describe '.new_from_url' do
```
edwardloveall, line 4:
```suggestion
  describe ".new_from_url" do
```

.sample.env
```diff
@@ -8,3 +8,5 @@ SMTP_ADDRESS=smtp.example.com
 SMTP_DOMAIN=example.com
 SMTP_PASSWORD=password
 SMTP_USERNAME=username
+ADMIN_USERNAME=admin
+ADMIN_PASSWORD=admin
```
edwardloveall, line 12: What do you think about putting these at the top to keep alphabetical order?

edwardloveall (approved):
Looks good! I put on my Professional Developer Hat™ and made a few comments which are all optional since this is a silly side project not a business.

edwardloveall:
Feel free to also address any or all of these in a follow up PR.

app/assets/stylesheets/admin.scss
gabebw: Sure, I'll factor this out. I chose this so we'd have a central column of some size with padding on the sides. Changing it won't break anything (but that could be clearer).

app/models/blocked_referrer.rb
gabebw: Yes, exactly. How about this?


> Create a blocked referrer that blocks everything from this site.
> If the URL's host is `sites.google.com`, it removes all info after `/sites/foobar`:
> sites.google.com/sites/foobar/baz/bat -> sites.google.com/sites/foobar
> This ensures that we do not block every single Google Site URL.
> Otherwise, it simply uses the site's host.

gabebw:
@edwardloveall Thank you for the insightful and comprehensive review! The code is definitely way better now.

app/models/blocked_referrer.rb
edwardloveall: Ship it

edwardloveall (approved):

PR #353, opened 2022-03-15
Allow editing URL in admin

gabebw:
#### Check out that "Edit" link

index

#### Ooooh, a form!

editing

### DIFF
--- a/app/views/admin/redirections/edit.html.erb
+++ b/app/views/admin/redirections/edit.html.erb
@@ -1,0 +1,8 @@
+<h1>Editing <code><%= @redirection.slug %></code></h1>
+
+<p>Original URL: <%= @redirection.original_url %></p>
+
+<%= simple_form_for [:admin, @redirection] do |form| %>
+  <%= form.input :url, label: "URL" %>
+  <%= form.submit %>
+<% end %>
--- a/spec/features/admin_edits_redirection_spec.rb
+++ b/spec/features/admin_edits_redirection_spec.rb
@@ -1,0 +1,24 @@
+require "rails_helper"
+
+RSpec.feature "Admin edits redirection" do
+  scenario "edits URL" do
+    new_url = "https://example.com/new"
+    url = "https://foobar.neocities.org"
+    original_url = "https://original.neocities.org"
+    slug = "binturong"
+    redirection = create(:redirection, slug:, url:, original_url:)
+    admin_login
+    visit admin_root_path
+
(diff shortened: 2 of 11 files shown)


app/views/admin/redirections/edit.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Editing <code><%= @redirection.slug %></code></h1>
+
+<p>Original URL: <%= @redirection.original_url %></p>
+
+<%= simple_form_for [:admin, @redirection] do |form| %>
+  <%= form.input :url, label: "URL" %>
+  <%= form.button :submit %>
```
edwardloveall, line 7: Hot Tip! You can do `form.submit` as well.

lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%# frozen_string_literal: true %>
+<%%= simple_form_for(@<%= singular_table_name %>) do |f| %>
```
edwardloveall, line 2: TIL about `<%%` and `<%%=` https://stackoverflow.com/a/40487347/638966

spec/features/admin_edits_redirection_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.feature "Admin edits redirection" do
+  scenario "edits URL" do
+    new_url = "https://example.com/new"
+    url = "https://foobar.neocities.org"
+    original_url = "https://original.neocities.org"
+    slug = "binturong"
+    redirection = create(:redirection, slug:, url:, original_url:)
```
edwardloveall, line 9: Is this a shortcut for `create(:redirection, slug: slug, etc)`? I've never seen blank keyword arguments when calling a method.

edwardloveall (approved):

lib/templates/erb/scaffold/_form.html.erb
gabebw: Actually, let's just remove this. We don't really use scaffolding.

spec/features/admin_edits_redirection_spec.rb
gabebw: So your first comment was correct!

As for your second comment: Maybe you didn't set the local variable name when testing in IRB, or maybe you ran it in an IRB that was older than v3.1.0?

PR #360, opened 2022-05-12
Disable redirection creation

edwardloveall:
We've been getting a lot of sites that aren't valid. Some seem like spam, some seem like bugs, but there are a lot of them. This makes it impossible to join automatically and adds instructions to contact us if a site would like to join.

instructions screen shot

### DIFF
--- a/app/assets/stylesheets/_pages.scss
+++ b/app/assets/stylesheets/_pages.scss
@@ -77,3 +77,13 @@ ul {
   margin: 0;
   padding: 0;
 }
+
+.annoucement {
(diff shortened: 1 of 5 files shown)


gabebw:
I think the instructions are great.

Why skip the tests? I think the tests are working fine, right?

edwardloveall:
They don't because I'm disabling the creation of redirections via the controller. We have a bunch of tests for that so those fail.

gabebw:
Ohhh, of course, right.

gabebw (approved):

PR #368, opened 2022-07-31
Display *original* URL in RSS feed

gabebw:
This will make it easier for us to find the links in Slack.

### DIFF
--- a/spec/requests/feed_requests_spec.rb
+++ b/spec/requests/feed_requests_spec.rb
@@ -15,9 +15,9 @@ RSpec.describe "Feed requests" do
 
       get feed_path
 
-      expect(xml["title"]).to eq("Hotline Webring")
-      expect(xml["updated"]).to eq(redirection.created_at.iso8601)
-      expect(xml["author"]["name"]).to eq("Gabe and Edward")
+      expect(xml.css("feed > title").text).to eq("Hotline Webring")
+      expect(xml.css("feed > updated").text).to eq(redirection.created_at.iso8601)
+      expect(xml.css("feed > author > name").text).to eq("Gabe and Edward")
     end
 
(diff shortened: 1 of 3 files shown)


spec/requests/feed_requests_spec.rb
```diff
@@ -15,9 +15,9 @@
 
       get feed_path
 
-      expect(xml["title"]).to eq("Hotline Webring")
-      expect(xml["updated"]).to eq(redirection.created_at.iso8601)
-      expect(xml["author"]["name"]).to eq("Gabe and Edward")
+      expect(xml.css("feed > title").text).to eq("Hotline Webring")
+      expect(xml.css("feed > updated").text).to eq(redirection.created_at.iso8601)
+      expect(xml.css("feed > author > name").text).to eq("Gabe and Edward")
```
edwardloveall, line 20: This is a little nicer, thanks. Did you also consider `dig`? Not that I'm requesting you switch to that. These are both better alternatives.

edwardloveall (approved):
Thank you!

spec/requests/feed_requests_spec.rb
gabebw: `dig` is for hashes, right? The issue is that with multiple `<link>` tags, `Hash.to_xml` creates an array such that `xml["link"]` contains an array with all of the link tag values.

Since I want to be able to differentiate between `link[rel=alternate]` and plain `link`, I switched to real XML parsing.

edwardloveall: Gotcha. You're right, the proper parsing is better.

FYI `dig` also works for arrays:

```ruby
[[1, 2, 3]].dig(0, 2)
# => 3
```

PR #371, opened 2022-09-24
Block subdomains of an existing redirection


gabebw:
If we already have `foo.com`, we now block `bar.foo.com` from joining. Unfortunately we have to do `Redirection.all.find` to do this check. It's not great, but it's probably not that bad in practice since we don't have thousands of Redirections.

If we could add arbitrary extensions to Postgres, I'd do it inside Postgres with https://github.com/petere/pguri and compare `uri_host` plus some string manipulation, but we can't do that.

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -2,6 +2,7 @@ class RedirectionsController < ApplicationController
   before_action :ensure_referrer_is_not_localhost
   before_action :ensure_referrer_is_not_blocked
   before_action :ensure_request_is_not_from_a_bot
+  before_action :ensure_request_is_not_a_subdomain_of_an_existing_site
 
   def next
     redirection = find_or_create_redirection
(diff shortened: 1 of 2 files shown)


edwardloveall (approved):

gabebw:
@edwardloveall While you're here...

I see that we now have 4 `before_action`s in this controller, and they're getting rather complex. What do you think about:

1. Moving all of the actions into a single class, like `RedirectionValidator`
2. Moving all of the actions into multiple classes (1 class per `before_action`) then composing them with `RedirectionValidator`

In particular, I'd like to skip all of these validations if we're simply looking up an existing slug, since we only need to validate when _creating_. This means we can skip the `Redirection.all.find` for 90% of requests. I know, it doesn't really affect performance, but it's bugging me.

I _could_ add `, unless: :creating_existing_slug?` to each `before_action`, but I think we share a horror of that DSL (and we are morally correct to do so).

So: extracting before_actions into classes, yay/nay? (In a separate PR, of course.)

edwardloveall:
Yeah, I'd agree. Maybe start with 1 and move to 2 if it still feels too crowded. And yeah, keep away from a web of callbacks.

PR #372, opened 2022-10-09
Help people who try to use an existing slug


gabebw:
## Clarify the instructions for joining
* Emphasize that they should not pick an existing slug
* Add angle brackets inside the URL so that if they try to copy-paste, it does not work.

CleanShot 2022-10-09 at 09 32 18@2x

## Show people a special page when they pick an existing slug

CleanShot 2022-10-09 at 09 30 39@2x

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -3,6 +3,7 @@ class RedirectionsController < ApplicationController
   before_action :ensure_referrer_is_not_blocked
   before_action :ensure_request_is_not_from_a_bot
   before_action :ensure_request_is_not_a_subdomain_of_an_existing_site
+  before_action :ensure_slug_matches_original_domain
 
   def next
     redirection = find_or_create_redirection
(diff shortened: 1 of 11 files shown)


app/controllers/redirections_controller.rb
```diff
@@ -100,7 +101,7 @@ def ensure_request_is_not_from_a_bot
   def ensure_request_is_not_a_subdomain_of_an_existing_site
     if referrer
       referrer_without_leading_subdomain = URI.parse(referrer).host.
-        sub(/^[^.]+\./, '')
+        sub(/^[^.]+\./, "")
```
edwardloveall, line 104: nice.

edwardloveall (approved):

app/controllers/redirections_controller.rb
gabebw: All credit to rufo 😄 https://github.com/ruby-formatter/rufo

edwardloveall:
We should definitely add this, but it may not fully solve the problem. Some people fill out one slug correctly and one incorrectly. They only need to click the correct one to join the 'ring.

gabebw:
But when someone clicks the bad one (perhaps the original owner), they'll see this message. And then hopefully they'll email us. Or their friend will click it, and tell the owner.

gabebw:
I think it's fine if they join with a good slug but have one broken link.

PR #373, opened 2022-10-09
Only validate new users when they try to join


gabebw:
Most of our `before_action`s are about validating *new* users. But most of our traffic is *existing* users who already have a `Redirection` and we just need to follow it.

For those existing users, validating the slug, referrer, etc is a waste of resources. Now we skip doing that work.

### DIFF
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -20,9 +20,8 @@ class RedirectionsController < ApplicationController
   private
 
   def find_or_create_redirection
-    redirection = Redirection.find_by(slug: params[:slug])
-    if redirection
-      redirection
+    if existing_redirection
+      existing_redirection
     else
       # Always log headers when possibly creating a new Redirection.
       # This ensures that if something goes wonky, we can look back through the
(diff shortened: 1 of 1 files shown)


app/controllers/redirections_controller.rb
```diff
@@ -20,9 +20,8 @@ def previous
   private
```
gabebw, line 20: This diff is best viewed with whitespace ignored: https://github.com/hotline-webring/hotline-webring/pull/373/files?diff=unified&w=1

app/controllers/redirections_controller.rb
```diff
@@ -73,33 +72,33 @@ def tagged_log(message)
...
+        redirect_to page_path("localhost")
+      end
     end
   end
 
   def ensure_referrer_is_not_blocked
-    if Redirection.exists?(slug: params[:slug])
-      # If the redirection already exists, let it through. We just want to
-      # prevent further redirections from being created.
-      return
```
gabebw, line 85: I replaced this block with `if creating_new_redirection?`, which does the same thing.

app/controllers/redirections_controller.rb
```diff
@@ -73,33 +72,33 @@ def tagged_log(message)
   end
 
   def ensure_referrer_is_not_localhost
-    if HostValidator.new(referrer).invalid?
-      redirect_to page_path("localhost")
+    if creating_new_redirection?
```
edwardloveall, line 75: If you like you could also put these in the `before_action`s:

```rb
before_action :ensure_referrer_is_not_localhost, if :creating_new_redirection?
```

edwardloveall (approved):

app/controllers/redirections_controller.rb
gabebw: Yeah, but:

1. I dislike the DSL for stuff that's already in Ruby
2. I want to move these to their own classes someday and it's a bit easier if everything is in the function already.

PR #387, opened 2023-02-06
Add API endpoints for our slack bot


gabebw:
The endpoints are:

```
POST  /api/redirections/:redirection_slug/blocks
POST  /api/redirections/:redirection_slug/unlinks
```

We use the slug because the Slack bot can only figure out the slug from the messages it sees in Slack. The slug is unique, so it shouldn't be an issue.

To see how the Slack bot works, see the (private) repo https://github.com/hotline-webring/sauron

### DIFF
--- a/app/controllers/api/blocks_controller.rb
+++ b/app/controllers/api/blocks_controller.rb
@@ -1,0 +1,11 @@
+class Api::BlocksController < ApiController
+  def create
+    redirection = Redirection.find_by!(slug: params[:redirection_slug])
+
+    if BlockedReferrer.block_and_unlink(redirection)
+      render json: { success: true }, status: 201
+    else
+      render json: { success: false }, status: 500
+    end
+  end
+end
--- a/.sample.env
+++ b/.sample.env
@@ -1,6 +1,7 @@
 # http://ddollar.github.com/foreman/
 ADMIN_PASSWORD=admin
 ADMIN_USERNAME=admin
+API_TOKEN=my-cool-api-token
 APPLICATION_HOST=localhost:3000
 ASSET_HOST=localhost:3000
 EXECJS_RUNTIME=Node
--- a/README.md
+++ b/README.md
@@ -64,6 +64,11 @@ To allow creating new redirections again, unset the variable:
 
     production config:unset DISALLOW_CREATING_NEW_REDIRECTIONS
 
+## API
+
+We have an API, oddly enough. It's used by our Slack bot, and we don't expect
+(or allow) anyone else to use it.
+
 ## Guidelines
 
 Use the following guides for getting things done, programming well, and
--- a/app/controllers/admin/blocks_controller.rb
+++ b/app/controllers/admin/blocks_controller.rb
@@ -1,9 +1,9 @@
 class Admin::BlocksController < AdminController
   def create
     redirection = Redirection.find(params[:redirection_id])
-    blocked_referrer = BlockedReferrer.new_from_url(redirection.url)
+    blocked_referrer = BlockedReferrer.block_and_unlink(redirection)
 
-    if blocked_referrer.save && redirection.unlink
+    if blocked_referrer
       flash[:success] = "Blocked and unlinked #{blocked_referrer.host_with_path}"
       redirect_to admin_redirections_path
     else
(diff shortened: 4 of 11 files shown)


spec/controllers/api_blocks_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+RSpec.describe Api::BlocksController do
```
edwardloveall, line 3: Maybe not important, but controller tests are no longer recommended (and I'm surprised this worked without the [rails-controller-testing](https://github.com/rails/rails-controller-testing) gem). It looks like you created a request spec in spirit anyway.

spec/controllers/api_blocks_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(response.status).to eq 401
+        expect(Redirection.find_by(slug: redirection.slug)).not_to be_nil
+        expect(BlockedReferrer.count).to eq 0
+      end
+    end
+
+    it "unlinks the redirection" do
+      redirection = create(:redirection, url: "https://foobar.neocities.org")
+
+      request.headers["Authorization"] = "Bearer #{ApiController::API_TOKEN}"
```
edwardloveall, line 20: How does this `API_TOKEN` work in testing? Does it grab the token from the .sample.env?

spec/controllers/api_unlinks_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe "POST :create" do
+    context "if API token is incorrect" do
+      it "returns a 401 status code" do
+        redirection = Redirection.find_by!(slug: "gabe")
+
+        post :create, params: { redirection_slug: redirection.slug }
+
+        expect(response.status).to eq 401
+        expect(Redirection.find_by(slug: redirection.slug)).not_to be_nil
+        expect(BlockedReferrer.count).to eq 0
```
edwardloveall, line 13: Optional:

```suggestion
        expect(BlockedReferrer).to be_empty
```

At least I think that will work.

edwardloveall (approved):
Woooo!

app/controllers/api/blocks_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Api::BlocksController < ApiController
+  def create
+    redirection = Redirection.find_by!(slug: params[:redirection_slug])
+    blocked_referrer = BlockedReferrer.new_from_url(redirection.url)
+
+    if blocked_referrer.save && redirection.unlink
```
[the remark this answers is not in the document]
gabebw: Sure, we can. I'll do the same in the Admin controller that does the same thing.

spec/controllers/api_blocks_controller_spec.rb
gabebw: It grabs it from `.env` (which must be created locally — it's gitignored). When Rails boots up, Dotenv kicks in and loads `.env`, and this spec uses a fully-booted Rails environment. So that constant is set.

spec/controllers/api_unlinks_controller_spec.rb
gabebw: Unfortunately it won't. It would have to be one of these:

```
expect(BlockedReferrer.any?).to be_true
```

```
# exists? does a fast count
expect(BlockedReferrer).not to_exist
```

I don't like either of these - `not to_exist` implies that the table is destroyed or something, and `any?` feels off.

What do you think?

spec/controllers/api_blocks_controller_spec.rb
gabebw: They are now request specs!

app/controllers/api/blocks_controller.rb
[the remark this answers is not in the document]
gabebw: Done in db1ab70f8ec632eb11567a47dd8f91289f1ca10b

spec/controllers/api_unlinks_controller_spec.rb
edwardloveall: totally fair. I think it could also be `expect(BlockedReferrer.count).to be_zero` but I'm just bike-shedding at this point. What you wrote initially is clear and fine.

spec/controllers/api_blocks_controller_spec.rb
edwardloveall: Oh, duh. I was thinking that `.env` didn't work in test mode or something, which is not true.

PR #394, opened 2023-05-29
Update from Rails 7.0.3.1 to 7.0.5

gabebw:
This is mostly just for security patches, e.g. https://github.com/advisories/GHSA-579w-22j4-4749, but also it's a good idea to periodically upgrade regardless.

Dependabot can't update Rails because it only updates one gem at a time and doesn't know that it needs to group Rails-related updates together.


edwardloveall (approved):
@gabebw merge

PR #396, opened 2023-06-10
Remove (most) references to heroku

edwardloveall:
We don't use heroku anymore. One of the references referenced rails 4.1. I figured it was safe to delete. There's also some references in the README and in `config/puma.rb`. I'm not sure what the equivalents are with fly.io though so I left em.

Please appreciate the branch name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,7 @@ Set up the repo:
 
 Run the app using [Heroku Local]:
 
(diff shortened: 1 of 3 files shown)


edwardloveall:
Whoops. We do use heroku.

gabebw:
HeYESku

PR #398, opened 2023-06-10
Note about what page to put your links on

edwardloveall:
Some people will put their links on a page that's pretty difficult to find. That's fine, but they should know that we will find it and redirect people there. This encourages them to put the links on a cool page.

### DIFF
--- a/app/views/homes/show.html.erb
+++ b/app/views/homes/show.html.erb
@@ -61,8 +61,14 @@
           <li>https://hotlinewebring.club/YOUR-SLUG/next</li>
           <li>https://hotlinewebring.club/YOUR-SLUG/previous</li>
           <li>
-            Please don't literally use YOUR-SLUG. It is already taken. Replace
+            <strong>A hot(line) tip</strong>: don't literally use YOUR-SLUG. It is already taken. Replace
             "YOUR-SLUG" with whatever you want your actual slug to be.
+          </li>
+          <li>
+            <strong>Another hot(line) tip</strong>: put your links on your coolest page. We look at
+            every site that joins and ensure visitors land on the page with the
+            webring links. This lets people click through every site in the
+            webring.
           </li>
         </ul>
       </li>


app/views/homes/show.html.erb
```diff
@@ -61,9 +61,14 @@
...
           <li>https://hotlinewebring.club/YOUR-SLUG/previous</li>
           <li>
-            Please don't literally use YOUR-SLUG. It is already taken. Replace
+            A hot(line) tip: don't literally use YOUR-SLUG. It is already taken. Replace
             "YOUR-SLUG" with whatever you want your actual slug to be.
           </li>
+          <li>
+            Another hot(line) tip: put your links on your coolest page. We look at
+            every site the joins and make sure that when someone visits you the
+            next and previous links will be on the first page they see.
```
edwardloveall, line 70: Open to different copy here

app/views/homes/show.html.erb
```diff
@@ -61,9 +61,14 @@
           <li>https://hotlinewebring.club/YOUR-SLUG/next</li>
           <li>https://hotlinewebring.club/YOUR-SLUG/previous</li>
           <li>
-            Please don't literally use YOUR-SLUG. It is already taken. Replace
+            A hot(line) tip: don't literally use YOUR-SLUG. It is already taken. Replace
```
gabebw, line 64: Hahaha

app/views/homes/show.html.erb
gabebw: Suggestion:

>Another hot(line) tip: put your links on your coolest page. We look at every site that joins and change the links (if necessary) to make sure that incoming webring links land on the page of your site with our webring links. This lets people click through every site in the webring.

edwardloveall: Nice. I took this and did even more edits. I don't think most people understand that we only get the host (without the path) so changing is kind of a weird concept. I tried to highlight the important bits. Still willing to take suggestions, though!

gabebw (approved):

PR #401, opened 2023-10-13
Bump Rails to the latest patch version


gabebw:
This fixes two security issues:

* https://github.com/advisories/GHSA-4g8v-vg43-wpgf
* https://github.com/advisories/GHSA-cr5q-6q9f-rq6q

Honestly, these almost certainly don't affect us, because we have no forms and take in almost zero user input, but our stellar security record must remain unblemished.


edwardloveall (approved):
Thanks!

PR #403, opened 2023-12-15
Allow us to close registration

edwardloveall:
We're overwhelmed with new registrations. This allow us to set the environment variable `CLOSED_FOR_REGISTRATION=1` which will allow all existing registrations to work, but redirect all new registrations to a "closed" page.

It also updates the homepage to show that we're closed so people can easily see before signing up.

Note that this is different from the
`DISALLOW_CREATING_NEW_REDIRECTIONS` config, which is more like a shadow ban when we're getting a bunch of spam. It does not show a message. Surprisingly different implementations!

Firefox Developer Edition


app/views/homes/show.html.erb
```diff
@@ -4,9 +4,18 @@
     Hotline Webring
     <a href="https://hotlinewebring.club/hotlinewebring/next">&rarr;</a>
   </h2>
+  <% if Rails.configuration.closed %>
+    <h2>Is Closed for Registration</h2>
```
gabebw, line 8: "Closed For Registration" is slightly awkward phrasing to me. What about "Is Not Accepting New Sites"?

Screenshot:
CleanShot 2023-12-15 at 11 03 45@2x

gabebw: Also: we use the `<h2>` and `<h3>` text here and in the `_closed` partial. Can you move it to I18n so we can do something like:

```erb
<%= t("closed_for_registration.subheader") %>
```

and

```erb
<%= t("closed_for_registration.existing_sites_still_work") %>
```

and use that in both places?

app/views/homes/show.html.erb
```diff
@@ -4,9 +4,18 @@
...
+    <h3>(But Existing Sites Still Work)</h3>
+  <% end %>
 </header>
 
 <main>
+  <% if Rails.configuration.closed %>
+    <%= render(partial: "pages/closed_text") %>
+    <p>We've kept the rest of this page intact for reference, and because we put a lot of work into it.</p>
```
gabebw, line 16:
```suggestion
    <p>We've kept the rest of this page intact, both for reference and because we put a lot of work into it.</p>
```

app/views/pages/_closed_text.html.erb
```diff
@@ -0,0 +1,4 @@
+<section class="text">
+  <p>Due to a frankly astonishing number of members in our silly webring, we're closing registrations for the moment. If we're being totally honest, the Ringleaders (Gabe and Edward) are also kinda tired of sifting through new and truly inventive web pages to find <code>next</code> and <code>previous</code> links and we wanted a break.</p>
```
gabebw, line 2:
```suggestion
  <p>Due to a frankly astonishing number of members in our silly webring, we're closing registrations for the moment. If we're being totally honest, the Ringleaders (Gabe and Edward) are also kinda tired of sifting through your wonderful and truly inventive web pages to find <code>next</code> and <code>previous</code> links and we wanted a break.</p>
```

spec/services/redirection_creation_spec.rb
```diff
@@ -45,4 +45,17 @@
       expect(result).to be_nil
     end
   end
+
+  context "when the site is closed" do
```
gabebw, line 49:
```suggestion
  context "when the 'ring is closed" do
```

Super small point, but I liked reading `'ring` in the spec above. It feels confidently flippant.

gabebw (commented):
Thank you for building this! I have some small nits, mostly around prose.

spec/services/redirection_creation_spec.rb
edwardloveall: confidently flippant is extremely on brand for you

gabebw: ❤️

gabebw (approved):

PR #420, opened 2024-10-19
Update dependencies with new script

edwardloveall:
I've been working on this script for a while on my own website. It updates gems that are returned by `bundle outdated --only-explicit` (just gems in the Gemfile, not dependencies) and then also checks [bundler-audit](https://github.com/rubysec/bundler-audit) for any remaining gem updates due to security vulnerabilities.

It's meant to be a local replacement for dependabot. I used it here to upgrade our gems.

If you'd rather stick with only dependabot, that's fine too. I can remove this.


gabebw (approved):
Neato!

PR #431, opened 2025-03-23
Tests were failing because the webring was closed


gabebw:
* Instead of relying on global `Rails.configuration`, pass in whether or not the webring is closed as an argument to `RedirectionCreation`
* Fix tests appropriately

### DIFF
--- a/app/services/redirection_creation.rb
+++ b/app/services/redirection_creation.rb
@@ -1,15 +1,16 @@
 class RedirectionCreation
-  def self.perform(referrer, slug)
-    new(referrer, slug).perform
+  def self.perform(referrer, slug, open:)
+    new(referrer, slug, open:).perform
   end
 
-  def initialize(referrer, slug)
+  def initialize(referrer, slug, open:)
     @referrer = referrer
     @slug = slug
+    @open = open
   end
 
   def perform
-    if Rails.configuration.closed
+    if !@open
       return ClosedRedirection.new
     end
 
--- a/app/controllers/redirections_controller.rb
+++ b/app/controllers/redirections_controller.rb
@@ -35,7 +35,11 @@ class RedirectionsController < ApplicationController
     if Rails.configuration.disallow_creating_new_redirections
       false
     else
-      RedirectionCreation.perform(referrer, params[:slug])
+      RedirectionCreation.perform(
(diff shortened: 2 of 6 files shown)


gabebw:
I got curious about why we didn't catch this before. I checked the original PR where we allowed closing registration, and the tests were passing then: https://github.com/hotline-webring/hotline-webring/pull/403

So really what's going on here is that I have `CLOSED_FOR_REGISTRATION=1` set in my local `.env` file. That makes the existing tests fail.

I could close this PR, because CI is passing (and I'll update my `.env`). However, I think this PR is still a good one:

1. I like explicitly passing in `closed:` to the service, instead of relying on global state.
2. We now have a bit more code coverage
  a. Coverage with closed registrations in `main`: 93.6% (`COVERAGE=1 CLOSED_FOR_REGISTRATION=1 rake; open coverage/index.html`)
  b. Coverage with open registrations in `main`: 96.51% (`COVERAGE=1 CLOSED_FOR_REGISTRATION=nope rake && open coverage/index.html`)
  c. Coverage with any (open _or_ closed) registrations in this branch: ✨ 96.52% ✨ (`COVERAGE=1 CLOSED_FOR_REGISTRATION=1 rake && open coverage/index.html`)

app/services/redirection_creation.rb
```diff
@@ -1,15 +1,16 @@
 class RedirectionCreation
-  def self.perform(referrer, slug)
-    new(referrer, slug).perform
+  def self.perform(referrer, slug, closed:)
+    new(referrer, slug, closed: closed).perform
```
edwardloveall, line 3: With newer ruby versions (3.1 I think?), you can just use:

```suggestion
    new(referrer, slug, closed:).perform
```

I care about a 1/5 about this, but thought you should know.

edwardloveall (approved):
Thanks for fixing this. Love 2 pass args directly.

I'm being a bit picky, but what do you think about switching to "open" or some other word that is closer to a positive/default happy path? I've found in the past that passing a boolean that does the opposite of the action gets harder to reason about e.g. `form.submit(cancel: true)` or `if !account.inactive` are kinda confusing when you could instead have `form.submit(confirm: true)` or `if account.active?`.

PR #432, opened 2025-03-23
Upgrade to Rails 8


gabebw:
* Opt in to every Rails 8 default config option
* Update rspec-rails to fix error message
  * This was happening with every test:
    ```
    NoMethodError:
    undefined method `fixture_path=' for class RSpec::ExampleGroups::[every group]
    ```

For a complete list of every `config.*` option, see here:  https://guides.rubyonrails.org/configuring.html

* I changed the Gemfile to point to Rails 8, then ran `rails app:update`
* The `public` files I accepted as-is. I'm fine using the generated files there, as we have been all along.
* `config/puma.rb` changes:
  * Remove `rackup DefaultRackup` since it's no longer needed and works fine without it
  * Per [Heroku](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#on-worker-boot): "Using `on_worker_boot` is no longer needed for Rails 5.2+ apps as forked connections will automatically re-connect."
  * Add PID file stuff from the generated Rails config; why not?
  * Get port from `ENV["PORT"]` explicitly (I assume it was implicit before); why not? (This was generated by Rails.)
* Expand `filter_parameter_logging` a bit, why not?
* `config/environments/development.rb`:
  * Set `config.active_record.query_log_tags_enabled = true`
  * It's the generated code; why not?
  * More information on this: https://api.rubyonrails.org/v7.1/classes/ActiveRecord/QueryLogs.html
  * `config.action_controller.raise_on_missing_callback_actions = true` seems like it is straightforwardly good and will catch some mistakes.
* `config/environments/production.rb`:
  * `enable_reloading = false` is the hot new way to say `cache_classes = true`
  * Sure, set far-future cache-control headers
  * Set `config.active_support.report_deprecations = false`, which is the default. I don't think we need to see deprecations in production.
  * Keep our existing behavior when calling `.inspect` on objects in production
  * `config.active_record.attributes_for_inspect = :all`


edwardloveall (approved):
Woo! 8️⃣

PR #438, opened 2025-06-27
Add "Last seen" column to admin interface

edwardloveall:
In order to help track sites that have gone defunct, we can track the last time they were "seen" by updating a `last_seen_at` column every time someone clicks their next or previous link.

Waterfox

I also updated the admin interface to actually be a grid instead of each line being it's own grid. Now all the columns line up.

### DIFF
--- a/app/controllers/admin/redirections_controller.rb
+++ b/app/controllers/admin/redirections_controller.rb
@@ -1,6 +1,11 @@
 class Admin::RedirectionsController < AdminController
+  DEFAULT_SORT_PARAMS = {"sort_key" => "next_id", "sort_dir" => "desc"}
+
   def index
-    @redirections = Redirection.order(created_at: :desc)
+    sort_by = DEFAULT_SORT_PARAMS.merge(sort_params)
+    order_by = { sort_by["sort_key"] => sort_by["sort_dir"] }
+
+    @redirections = Redirection.order(order_by)
   end
 
   def edit
@@ -23,5 +28,9 @@ class Admin::RedirectionsController < AdminController
 
   def redirection_params
     params.require(:redirection).permit(:url)
+  end
+
+  def sort_params
+    params.permit(:sort_key, :sort_dir)
   end
 end
--- a/app/controllers/admin/redirections_controller.rb
+++ b/app/controllers/admin/redirections_controller.rb
@@ -3,9 +3,7 @@ class Admin::RedirectionsController < AdminController
 
   def index
     sort_by = DEFAULT_SORT_PARAMS.merge(sort_params)
-    order_by = { sort_by["sort_key"] => sort_by["sort_dir"] }
-
-    @redirections = Redirection.order(order_by)
+    @redirections = Redirection.order_nulls_last(sort_by)
   end
 
   def edit
--- a/app/views/admin/redirections/index.html.erb
+++ b/app/views/admin/redirections/index.html.erb
@@ -1,36 +1,62 @@
-<div class="header-row">
-  <div class="redirection">
-    <div>Slug</div>
-    <div class="url">URL</div>
-    <div class="original-url">Original URL</div>
-    <div class="edit"></div>
-    <div class="block-button"></div>
-    <div class="unlink-button"></div>
-  </div>
-</div>
+<h2>Sort by</h2>
+<section class="sort">
+  <%= link_to(
+    "Ring order",
+    admin_redirections_path(sort_key: "next_id", sort_dir: "desc")
+  ) %>
+  <%= link_to(
+    "Most recently seen",
+    admin_redirections_path(sort_key: "last_seen_at", sort_dir: "desc")
+  ) %>
+  <%= link_to(
+    "Least recently seen",
+    admin_redirections_path(sort_key: "last_seen_at", sort_dir: "asc")
+  ) %>
+</section>
 
-<% @redirections.each do |redirection| %>
-  <div class="redirection" data-id="<%= redirection.id %>">
-    <div>
-      <label>Slug</label>
-      <span class="content"><%= redirection.slug %></span>
-    </div>
-    <div class="url">
-      <label>URL</label>
-      <span class="content"><%= link_to redirection.url, redirection.url %></span>
-    </div>
-    <div class="original-url">
-      <label>Original URL</label>
-      <%= link_to redirection.original_url, redirection.original_url %>
-    </div>
-    <div class="edit">
-      <%= link_to "Edit", edit_admin_redirection_path(redirection) %>
-    </div>
-    <div class="block-button">
-      <%= button_to "Block and unlink", admin_redirection_blocks_path(redirection) %>
-    </div>
-    <div class="unlink-button">
-      <%= button_to "Unlink", admin_redirection_unlinks_path(redirection) %>
+<h2>Redirections</h2>
+<section class="redirections">
+  <div class="header-row">
+    <div class="redirection">
+      <div>Slug</div>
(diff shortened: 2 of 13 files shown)


app/controllers/admin/redirections_controller.rb
```diff
@@ -1,6 +1,9 @@
 class Admin::RedirectionsController < AdminController
+  DEFAULT_SORT_PARAMS = {"sort_key" => "next_id", "sort_dir" => "desc"}
+
   def index
-    @redirections = Redirection.order(created_at: :desc)
+    order_by = [DEFAULT_SORT_PARAMS.merge(sort_params).values].to_h
```
gabebw, line 5: This is a little tricky. I had to paste it into `irb` to see what it was doing. How about this?

```ruby
sort_by = DEFAULT_SORT_PARAMS.merge(sort_params)
order_by = { sort_by["sort_key"] => sort_by["sort_dir"] }
```

app/views/admin/redirections/index.html.erb
```diff
@@ -1,36 +1,62 @@
...
-  </div>
-</div>
+<h2>Sort by</h2>
+<section class="sort">
+  <%= link_to(
+    "Ring order",
+    admin_redirections_path(sort_key: "next_id", sort_dir: "desc")
+  ) %>
+  <%= link_to(
+    "Last seen",
```
gabebw, line 8: Since "Last seen" could mean it's sorted either ascending or descending, how about "Most recently seen first"?

app/views/admin/redirections/index.html.erb
```diff
@@ -1,36 +1,62 @@
...
-    <div class="block-button">
-      <%= button_to "Block and unlink", admin_redirection_blocks_path(redirection) %>
-    </div>
-    <div class="unlink-button">
-      <%= button_to "Unlink", admin_redirection_unlinks_path(redirection) %>
+<h2>Redirections</h2>
+<section class="redirections">
+  <div class="header-row">
+    <div class="redirection">
+      <div class="slug">Slug</div>
```
gabebw, line 21: Since it's a grid, does this need to have the classes? Does it need to be nested in a `.redirection`? Could it be this?

```erb
<section class="redirections">
  <div class="header-row">
    <% ["Slug", "URL", "Original URL", "Last seen", "", "", ""].each do |column| %>
      <div><%= column %></div>
    <% end %>
  </div>
<div>
```

db/migrate/20250627212857_add_last_seen_to_redirections.rb
```diff
@@ -0,0 +1,10 @@
+class AddLastSeenToRedirections < ActiveRecord::Migration[8.0]
+  def up
+    add_column :redirections, :last_seen_at, :datetime
+    Redirection.update_all(last_seen_at: Time.zone.now)
```
gabebw, line 4: Hmm, this isn't quite true, and it might give a false sense of when they were last seen. I think it should either be:

* `NULL` (and fall back to `created_at` in the view, as you wrote)
* Or set it to `created_at` for each one, make it `null: false`, and update it on creation

I think I prefer to set it to `NULL`, which is more accurate: we don't really have a value for it.

db/schema.rb
```diff
@@ -10,15 +10,14 @@
...
-ActiveRecord::Schema.define(version: 2020_07_03_012614) do
-
+ActiveRecord::Schema[8.0].define(version: 2025_06_27_212857) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "plpgsql"
+  enable_extension "pg_catalog.plpgsql"
 
   create_table "blocked_referrers", force: :cascade do |t|
     t.string "host_with_path", null: false
-    t.datetime "created_at", precision: 6, null: false
```
gabebw, line 20: Do you know why the schema changed these things (besides the `last_seen_at`)? Is it just part of it being an `[8.0]` schema now?

edwardloveall: Seems like it: https://stackoverflow.com/a/71482301/638966

This is the first migration in 5 years so it makes sense we've got some changes.

db/migrate/20250627212857_add_last_seen_to_redirections.rb
edwardloveall: I forgot I added that fallback logic for `NULL`. I'll do your second option.

app/views/admin/redirections/index.html.erb
edwardloveall: You're totally right it doesn't need the classes. I don't love the loop, though, especially with the empty strings. I'll just leave them their literal `div`s.

app/controllers/admin/redirections_controller.rb
edwardloveall: Oh yeah, that's much better.

app/views/admin/redirections/index.html.erb
```diff
@@ -1,36 +1,62 @@
...
-  </div>
-</div>
+<h2>Sort by</h2>
+<section class="sort">
+  <%= link_to(
+    "Ring order",
+    admin_redirections_path(sort_key: "next_id", sort_dir: "desc")
+  ) %>
+  <%= link_to(
+    "Most recently seen first",
```
gabebw, line 8: Hmm, how about "Most recently seen" here and "Least recently seen" below? Sorry, this is nitpicky.

app/views/admin/redirections/index.html.erb
```diff
@@ -1,36 +1,62 @@
...
-    </div>
-    <div class="block-button">
-      <%= button_to "Block and unlink", admin_redirection_blocks_path(redirection) %>
-    </div>
-    <div class="unlink-button">
-      <%= button_to "Unlink", admin_redirection_unlinks_path(redirection) %>
+<h2>Redirections</h2>
+<section class="redirections">
+  <div class="header-row">
+    <div class="redirection">
```
gabebw, line 20: Does this need to be here? Could it just be wrapped in a `.header-row` without `.redirection`?

db/migrate/20181220204317_add_original_url_to_redirection.rb
```diff
@@ -1,11 +1,5 @@
 class AddOriginalUrlToRedirection < ActiveRecord::Migration[5.2]
-  def up
+  def change
     add_column :redirections, :original_url, :text
-    update 'UPDATE redirections SET original_url = url'
```
gabebw, line 4: Hmm, why did this file need to change?

edwardloveall: Oof. Because I wasn't paying enough attention. Thank you.

app/views/admin/redirections/index.html.erb
edwardloveall: You lose the bottom margin because `.header-row` is `display: contents` which won't show margin, but it's fine.

gabebw: Ah, that's fine to keep as-is then.

db/migrate/20250627212857_add_last_seen_to_redirections.rb
edwardloveall: I will say this has a odd side-effect that if it's `nil` it gets sorter before others that have been seen. I'll see if I can fix that.

edwardloveall: Tried to solve it in 2cadbca Let me know if you can think of a better API though.

gabebw (approved):
Looks good! Thanks for building this.

PR #452, opened 2026-02-12
Use plain CSS and Sprockets (no SCSS)

gabebw:
This is a followup to the discussion about plain CSS here: https://github.com/hotline-webring/hotline-webring/pull/451#issuecomment-3888151310

This PR removes all Sass from our pipeline and uses plain CSS. (~~It still uses Sprockets to minify the CSS and combine it into one file at the end; we could remove Sprockets too and use `@import`.~~ Update: we do this now in 5be52a5e769592c422a9574f32100c7c827d0ef5.)

<strike>Honestly, I'm not sold on vanilla CSS!

1. I like the Sass variable syntax more: `$foo` is just more pleasant than `var(--foo)`.
2. While it's neat that you can do HSL calculations, I think `colors.adjust` is more readable.
3. We don't get that much for switching (besides a warm feeling): Sprockets is still there, and we still run it with `rails s`. It's not really that much more simpler.
</strike>

I'm in, let's use as much native browser functionality as we can.


gabebw:
Good point about tooling. I will think on it.

Re: HSL, yes, your relative colors link is what I meant. That's what I'm doing on [this line](https://github.com/hotline-webring/hotline-webring/pull/452/changes#diff-8af31a076802d418016ec66aa863fbab7871da0d062852d61a02f304ba6d6193R9):

```css
--dark-pink: hsl(from var(--hotline-bling-pink) h s calc(l - 40));
```

I'm not sure what [the Codepen](https://codepen.io/edwardloveall/pen/raLPNyd) is showing that is different from that?

edwardloveall:
> I'm not sure what [the Codepen](https://codepen.io/edwardloveall/pen/raLPNyd) is showing that is different from that?

Oh, yeah sorry. That's the same thing. I saw "draft" pr and didn't peek at the code, and I had already made the codepen. I'll just let you work 🙂

gabebw:
After thinking about it, I'm convinced: plain CSS is very much in the spirit of HLWR. I'm going to merge in a few days; any comments before I do?

edwardloveall (approved):
This is great! Thanks for converting everything.

Gemfile
```diff
@@ -17,6 +17,7 @@ gem "puma", "~> 7"
 gem "rack-canonical-host"
 gem "rails", "~> 8.1.0"
 gem "simple_form"
+gem "sprockets-rails"
```
gabebw, line 20: Rails requires an asset pipeline to exist, otherwise we get errors about `config.assets` being absent. So we can't entirely remove Sprockets from this project.

gabebw:
@edwardloveall I'm on board with this PR! I reduced our reliance on Sprockets even further in 5be52a5e769592c422a9574f32100c7c827d0ef5 — check it out.

app/assets/stylesheets/_variables.css
```diff
@@ -0,0 +1,19 @@
...
+  --dark-pink: hsl(from var(--hotline-bling-pink) h s calc(l - 40));
+  --pastel-yellow: #fff3cd;
+  --pastel-red: #fdaaaa;
+  --pastel-blue: #cde4ff;
+  --pastel-green: #d5ecdb;
+  --dark-red: #ef5350;
+  --border-color: hsl(from var(--dark-pink) h s calc(l - 10));
+
+  /* We picked a size so that it didn't span the entire page */
+  --admin-max-width: 80%;
```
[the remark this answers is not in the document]
gabebw: I'll keep it as-is, just for the sake of changing fewer things.

app/assets/stylesheets/_wide-screens.css
```diff
@@ -0,0 +1,13 @@
+@import url("_variables.css");
```
edwardloveall: A note about `@import` is that files are [not downloaded in parallel](https://stackoverflow.com/questions/10036977/best-way-to-include-css-why-use-import), so using `@import` a lot can really slow down the load time. It _probably_ does not matter to us, but we may want to just move all these to be `<link>` elements in the `<head>` if it does.

app/assets/stylesheets/admin.css
```diff
@@ -100,35 +99,35 @@ h2 {
...
-      3fr // Original URL
-      2fr // Last seen
-      1fr // edit
-      repeat(2, 2fr); // buttons
+      1fr /* Slug */
+      3fr /* URL */
+      3fr /* Original URL */
+      2fr /* Last seen */
+      1fr /* edit */
+      repeat(2, 2fr); /* buttons */
```
edwardloveall, line 132: You can also name these using css, if you'd rather:

```suggestion
      [slug] 1fr
      [url] 3fr
      [original-url] 3fr
      [last-seen] 2fr
      [edit] 1fr
      [buttons] repeat(2, 2fr); /* buttons */
```

edwardloveall (approved):

app/assets/stylesheets/_wide-screens.css
gabebw: Oh, I didn't realize it wasn't in parallel. I explored moving it all to `stylesheet_link_tag`, but Sprockets' require felt like the most ergonomic version after all, so we're back to that.

app/assets/stylesheets/admin.css
gabebw: Since we don't actually use the named grid slots, I'll keep these as comments.

PR #464, opened 2026-03-24
Upgrade to Ruby v4

gabebw:
Here's a list of the changes in 4.0.0: https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/

There's not much that affects us besides ZJIT, which gives better performance (but we are fast enough already).


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby "3.3.5"
+ruby "4.0.1"
```
edwardloveall: [Hot tip](https://bundler.io/man/gemfile.5.html#VERSION-required-):

```suggestion
ruby ".ruby-version"
```

Gemfile
```diff
@@ -30,6 +30,7 @@ end
 group :test do
   gem "capybara", "~> 3.0"
   gem "launchy"
+  gem "mutex_m"
```
edwardloveall: What's up with this?

edwardloveall (approved):

Gemfile
edwardloveall: Oh, it's for CI. Sure.

edwardloveall: We can remove capybara-mechanize if we want. I should probably just remove all the link audit code.

Gemfile
gabebw: Nice!

